PR #47, opened 2013-01-11
show a progress indicator while scanning

xrstf:
I use Satis to scan 60+ repositories in one go. It's hard to tell how far I already got, so I hacked together this simple IO wrapper that prepend the current and total number of repositories. So the output looks like this:

```
Q:\satis\bin>php compile
Scanning packages
[ 8/61] Reading composer.json of webvariants/developer-utils (v5.1.1)
Writing web view

```

I don't really expect this to be merged, as I'm sure there are other usecases where my code terribly fucks up. But maybe it's an inspiration :-)


alcohol:
Closing this due to lack of activity and conflicting state. Please do feel free to re-open and update this PR if said functionality is still desired. :-)

PR #110, opened 2014-01-10
Partial update

### LINKED ISSUE #40: Selective update
Currently we are using Satis to mantain our private package repositories. We have configured the Github hooks to update automatically when pushing. As we have many packages, this process take longer than expected (as is updates everything and not the just updated package).

Is there any way to tell Satis to update a specifig package (or packages)?. This way we could save a lot of time (and why not, Github API calls).

Thanks in advance.

jkufner:
This updates and enhances oker1's pull request #84.

Repository configuration from satis.json is now used. URL is used only to find appropriate configuration in satis.json. If URL is not found, an exception is thrown.

It also renames --repository option to --repository-url to better describe expected value.

It performs full update if cache file is not found.

Fixes #40.


mathroc:
:+1:

PR #160, opened 2014-07-06
Added name-match and latest-only option

bart:
Should fix issues #129, #139 and #158

PR #178, opened 2014-10-12
Add limits to which dependencies get automatically pulled in.

### LINKED ISSUE #158: Problem with nested non-available requirements
Hi everybody, I have a question about the require-dependencies flag. When I set it to true, it seams that satis loads dependencies to an unlimited depth. That means it loads dependencies, which are not needed by the package!

Example: codeception/codeception

{
    "name": "local repo",
    "homepage": "http://localrepo.local/packages",
    "repositories": [
        { "type": "composer", "url": "https://packagist.org" }
    ],
    "require": {
        "codeception/codeception": "~2.0"
    },
    "archive": {
        "directory": "dist",
        "skip-dev": true
    },
    "require-dependencies": true
}

When I load this directly via composer, it would never try to load sirsquall/guzzle as a dependency, because this seems to be an old one. The question is: Why does satis loads this as a dependency and how can I prevent it?

sirsquall/guzzle isn't available at github and so I get an error for a package, I don't need. Do you have any solution for that?

Thanks so much!

Leftbrained:
Adds two new keys to the satis.json file format. Conflicts with #160. Seems to fix #158.

* **dependency-minimum-stability** Used to specifiy the minimum stability for dependencies included in the require-dependencies and require-dev-dependencies options independently of the dependencies in the satis.json file.

Dependencies must satisfy both minimum-stability and dependency-minimum-stability.

* **only-latest-dependencies** Used to choose only the latest dependency for each package.

It will only ever install _one_ package per name, unless that package is listed in the satis.json file.

The first commit is a mild code refactor to put the select package functionality into a portable closure, so we don't have to maintain that functionality in two places.

It adds a simple stability check for _dependency-minimum-stability_.

For _only-latest-dependencies_ Instead of one giant pass to pull in packages from links it does them in chunks. The first pass is the root links, the rest are dependencies. Each pass does the following:
 1. Find all packages that match link and stability requirements. Root links get selected. Deps get stored in a  buffer to determine the latest at the end of the pass. 
 2. Loop through each package name found in dependencies and select the latest package version for each.

### DIFF
--- a/src/Composer/Satis/Command/BuildCommand.php
+++ b/src/Composer/Satis/Command/BuildCommand.php
@@ -144,7 +144,7 @@ EOT
         }
 
         $minimumStability =  isset($config['minimum-stability']) ? $config['minimum-stability'] : 'dev';
-        $dependencyMinimumStability =  isset($config['dependency-minimum-stability']) ? $config['dependency-minimum-stability'] : $minimumStability;
+        $dependencyMinimumStability =  isset($config['dependency-minimum-stability']) ? $config['dependency-minimum-stability'] : 'dev';
 
         if (!$outputDir = $input->getArgument('output-dir')) {
             $outputDir = isset($config['output-dir']) ? $config['output-dir'] : null;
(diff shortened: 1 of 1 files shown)


tralafiti:
Any news here? `only-latest-dependencies` would be really appreciated.

alcohol:
This PR is in a conflicting state. Do you wish to update it, or can it be closed?

PR #185, opened 2014-11-21
Purge command

caouecs:
Hello,
we created the purge command to delete all packages not needed by our new versions, and our tests. It based on the last archive file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,16 @@ Usage
 - Install satis: `php composer.phar create-project composer/satis --stability=dev --keep-vcs`
 - Build a repository: `php bin/satis build <composer.json> <build-dir>`
 
-Read the more detailed instructions in the 
+Read the more detailed instructions in the
 [documentation](http://getcomposer.org/doc/articles/handling-private-packages-with-satis.md).
+
+Purge
+-----
+
+If you choose to archive packages in your server, you can have useless files.  
+With the `purge` command, you delete these files.
+
+    php bin/satis purge <composer.json> <build-dir>
 
 Updating
 --------
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ With the `purge` command, you delete these files.
 
     php bin/satis purge <composer.json> <build-dir>
 
+ > Note: be careful if you reference your archives in your lock file.
+
 Updating
 --------
 
--- a/composer.lock
+++ b/composer.lock
@@ -12,12 +12,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/composer.git",
-                "reference": "fc57c97f9f6f1a309dd51a9312218d052b8940c7"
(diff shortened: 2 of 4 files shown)


raistlin:
Any chance this will be merged?

README.md
```diff
@@ -13,9 +13,17 @@ Usage
...
 
-Read the more detailed instructions in the 
+Read the more detailed instructions in the
 [documentation](http://getcomposer.org/doc/articles/handling-private-packages-with-satis.md).
 
+Purge
+-----
+
+If you choose to archive packages in your server, you can have useless files.  
+With the `purge` command, you delete these files.
```
till: :+1: 

I think this is a tricky feature to add.

GromNaN:
:+1:

Divi:
:+1:

alcohol:
How about adding an option/flag to specify how many builds to keep, e.g. to keep the last 3.

ro0NL:
@alcohol any updates on this? I think it's definitely a nice-to-have feature and should be reviewed/merged or at least decline the pull request otherwise...

alcohol:
There you go.

ro0NL:
That was fast :astonished: Thanks.

ro0NL:
@alcohol im not sure we're going to run this in production. Im testing it locally and everything got removed from `dist/` ;-)

Im debugging it right now.. and as far as I can see the regex at https://github.com/composer/satis/blob/master/src/Composer/Satis/Command/PurgeCommand.php#L100,L110 is totally failing :exclamation: 

Our results:
```php
$needed = array('vendor-some-bundle');
$files = array('vendor-some-bundle-v1.0.0-XXXXXX.zip')
```

So basicly the regex is applied as follow https://regex101.com/r/lF5mI7/1

I.e. `$matches[1]` is set to `vendor-some` whish is not in the `$needed` array, so unlink is called. However it actually "needed".

So the regex vs. package naming is problematic due dashes I guess. Also it has no start/end modifiers (`^...$`) which can be a potential risk/another false positive.

Im testing our usecase, and planning to fix it at regex-level.

For your information ;-)

ro0NL:
@alcohol following regex seems OK: `/^(.+)-(?:[^-]+)-(?:[^-]+)\.(?:.+)$/`

https://regex101.com/r/eW2aY8/1

Can u fix this in master or should i go with a PR?

ro0NL:
Also `include/` is never cleaned up.. which is a minor for now. I guess these can all be removed except the one referenced in `package.json`, right?

yoshz:
Please keep in mind that when you purge the dist files, any references in composer.lock of your project will still point to those files!

alcohol:
@ro0NL I might have a looksie tomorrow. But a PR is always welcome, so if you have the time, please feel free to submit one.

PR #197, opened 2015-01-11
Added ability to support repositories which have more than one package. 

andkirby:
We have a repository which has several packages (modules).
 We use following naming approach to separate them:
 ModuleName/master
 ModuleName/develop
 ModuleName/feature/some-name
 ModuleName/v1.0.0
 AnotherModuleName/master
 AnotherModuleName/develop
 AnotherModuleName/feature/some-name
 AnotherModuleName/v1.0.0

So, it will create package with different package names:
 vendorname/reponame-modulename
 vendorname/reponame-anothermodulename
 (seem need to separate words with "-")

Problem on the composer side:
composer will create huge repository each time. (composer uses `git fetch`)
(let's admit that we have 20 extension by 5MB. So if we install all via composer we will have at least 20 repositories by 100 MB (5\*20=100). In total 100\*20=2GB)

What can solve it. We could implement composer plugin which 
- will create the one GIT repository for all packages
- or implement functionality to make "dist" with repository autoarchives like https://github.com/andkirby/tar-composer-installer/archive/master.zip
(also needed for GitLab)

### DIFF
--- a/src/Composer/Satis/Composer/Factory.php
+++ b/src/Composer/Satis/Composer/Factory.php
@@ -1,0 +1,31 @@
+<?php
+namespace Composer\Satis\Composer;
+
+use Composer\Config;
+use Composer\EventDispatcher\EventDispatcher;
+use Composer\IO\IOInterface;
+
+/**
+ * Class Factory
+ *
+ * \Composer\Factory was extended to add custom repository class
+ *
+ * @package Composer\Satis\Composer
+ */
+class Factory extends \Composer\Factory
+{
+    /**
+     * Added additional required VCS type "vcs-namespace"
+     *
+     * @param IOInterface $io
+     * @param Config $config
+     * @param EventDispatcher $eventDispatcher
+     * @return \Composer\Repository\RepositoryManager
+     */
+    protected function createRepositoryManager(IOInterface $io, Config $config, EventDispatcher $eventDispatcher = null)
+    {
+        $rm = parent::createRepositoryManager($io, $config, $eventDispatcher);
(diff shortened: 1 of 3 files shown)


Cog-g:
Hi. Sorry if I missed something but...why can't you use branch/tag for versions?

src/Composer/Satis/Composer/Factory.php
```diff
@@ -0,0 +1,31 @@
...
+     * @param EventDispatcher $eventDispatcher
+     * @return \Composer\Repository\RepositoryManager
+     */
+    protected function createRepositoryManager(IOInterface $io, Config $config, EventDispatcher $eventDispatcher = null)
+    {
+        $rm = parent::createRepositoryManager($io, $config, $eventDispatcher);
+        $rm->setRepositoryClass('vcs-namespace', 'Composer\Satis\Repository\VcsNamespaceRepository');
+        return $rm;
+    }
+}
```
GrahamCampbell: files must end in a single new line

src/Composer/Satis/Repository/VcsNamespaceRepository.php
```diff
@@ -0,0 +1,293 @@
...
+     *
+     * @param string $branch
+     * @return string
+     */
+    protected function _getBranchNamespace($branch)
+    {
+        $arr = explode('/', $branch);
+        return array_shift($arr);
+    }
+}
```
GrahamCampbell: same

till:
cc @stof — Can we use PHP-cs-fixer to enforce a certain "rule set" easily?

andkirby:
@GrahamCampbell, I don't know maybe I missed something.. But, look..
I have namespaces in tags, so satis cannot recognize them.

PR #199, opened 2015-01-14
Showing raw name

Inscure:


bishopb:
Shouldn't <title> also have a similar change?
https://github.com/composer/satis/blob/master/views/index.html.twig#L6

PR #213, opened 2015-03-20
Add abandon notices to default twig template

SvenRtbg:
With the ability to tag projects as abandoned in Composer, this information should be shown in Satis as well. I tried a simple approach, both enabling this info for the highest package version in a prominent place, as well as for every version in case of the most recent version(s) not being abandoned. There is the ability to style these infos via a new CSS class "abandoned".


JamesRezo:
This can be closed in favor of #220

PR #218, opened 2015-04-18
improve search performance

Flyingmana:
each toggle() of a package triggers a render of the browser content,
to reduce the number of renders, we hide the element containing all
packages during the filtering.

with the big number of packages we have at http://packages.firegento.com/ a search can need  >30s.
JS profiling showed, the JS is only busy ~2-3s and idles during the rest.

with this patch the rendering time for a search was reduced to just a few seconds.


Flyingmana:
the satis at http://packages.firegento.com/ now include this changes, so you can test them there.

PR #224, opened 2015-06-03
Remove unnecessary pass-by-reference

joshdifabio:
`BuildCommand::dumpDownloads()` uses pass-by-reference for the `$packages` param. Presumably this was once necessary but there doesn't appear to be any reason for it now. Am I missing something?


stof:
:+1:

PR #225, opened 2015-06-03
Randomise package download order

joshdifabio:
Initial builds can take a very long time in Satis when there are many dependencies. This PR takes a pragmatic approach to speeding this up by making it practical to run many `satis build` commands simultaneously using the same configuration: The download order of the packages is simply randomised to reduce the likelihood of multiple builds attempting to download the same packages at the same time. We have used this approach internally a number of times and it's helped a lot.

This PR includes the changes made in #224.

### DIFF
--- a/src/Composer/Satis/Command/BuildCommand.php
+++ b/src/Composer/Satis/Command/BuildCommand.php
@@ -327,7 +327,7 @@ EOT
 
     /**
      * @param array           $config   Directory where to create the downloads in, prefix-url, etc..
-     * @param array           $packages Reference to packages so we can rewrite the JSON.
+     * @param array           $packages
      * @param InputInterface  $input
      * @param OutputInterface $output
      * @param string          $outputDir
@@ -335,7 +335,7 @@ EOT
      *
      * @return void
      */
-    private function dumpDownloads(array $config, array &$packages, InputInterface  $input, OutputInterface $output, $outputDir, $skipErrors)
+    private function dumpDownloads(array $config, array $packages, InputInterface  $input, OutputInterface $output, $outputDir, $skipErrors)
     {
         if (isset($config['archive']['absolute-directory'])) {
             $directory = $config['archive']['absolute-directory'];
--- a/src/Composer/Satis/Command/BuildCommand.php
+++ b/src/Composer/Satis/Command/BuildCommand.php
@@ -364,11 +364,14 @@ EOT
 
         $archiveManager->setOverwriteFiles(false);
 
+        shuffle($packages);
         /* @var \Composer\Package\CompletePackage $package */
-        foreach ($packages as $name => $package) {
+        foreach ($packages as $package) {
             if ('metapackage' === $package->getType()) {
                 continue;
             }
+
+            $name = $package->getName();
 
             if (true === $skipDev && true === $package->isDev()) {
                 $output->writeln(sprintf("<info>Skipping '%s' (is dev)</info>", $name));


stof:
What is the goal of shuffling packages to make builds faster ?

joshdifabio:
Sorry, the description in the OP is a bit poor.

Consider the following minimized Satis config:

```json
{
    "require": {
        "my/package1": "*",
        "my/package2": "*",
        "my/package3": "*"
    }
}
```

Now, imagine we trigger two Satis builds for the first time:

```
satis build my-config.json & # process 1
satis build my-config.json & # process 2
# process 1 finds that my/package1 @ 1.0.0 is missing
# process 1 begins to download my/package1 @ 1.0.0
# process 2 finds that my/package1 @ 1.0.0 is missing
# process 2 begins to download my/package1 @ 1.0.0
# process 1 finds that my/package1 @ 1.0.1 is missing
# process 1 begins to download my/package1 @ 1.0.1
# process 2 finds that my/package1 @ 1.0.1 is missing
# process 2 begins to download my/package1 @ 1.0.1
...
```

Generally speaking, you don't get any benefit from running Satis in parallel because all the builds try and download the same files at the same time. Randomising the order helps with this.

```
satis build my-config.json & # process 1
satis build my-config.json & # process 2
# process 1 finds that my/package2 @ 2.3.5 is missing
# process 1 begins to download my/package2 @ 2.3.5
# process 2 finds that my/package1 @ 1.6.0 is missing
# process 2 begins to download my/package1 @ 1.6.0
# process 1 finds that my/package1 @ 1.0.1 is missing
# process 1 begins to download my/package1 @ 1.0.1
# process 2 finds that my/package1 @ 1.2.0 is missing
# process 2 begins to download my/package1 @ 1.2.0
...
```

stof:
oh, so the shuffling means that the second process will be likely to reuse the cache created in the first process. this looks interesting.

btw, the fact that shuffle takes an array by reference means that you get a [reference mismatch](http://jpauli.github.io/2014/06/27/references-mismatch.html) on the array, requiring to duplicate it in memory

joshdifabio:
Yes, it will be duplicated in memory, but as it's just an array of object references I figured it wasn't a big deal. Do you think there's a better approach?

naderman:
Indeed I think that's fine, and I don't see an issue with shuffling them, so think this is fine and can be merged. Not sure if there is a place where the order matters?

stof:
@naderman I don't think so. The order matters for the dumped packages.json (to make its hash deterministic) but it is not impacted by the shuffling done here as it is only shuffled for the dumping of archives.

joshdifabio:
One thing I should point out is that during large builds (many packages downloaded) it will become harder for the user to gauge progress. Assume I need to grab all versions of `my/package` and I know the versions are `1.0`, `1.1`, `1.2`, `2.0` and `2.1`. Instead of seeing:
```
Dumping 'my/package-1.0.0.0'
Dumping 'my/package-1.1.0.0'
Dumping 'my/package-1.2.0.0'
Dumping 'my/package-2.0.0.0'
Dumping 'my/package-2.1.0.0'
```
... I will instead see the version numbers in a random order. When there are many versions to download it will become harder to track the progress. If this PR gets merged, it might make sense for us to make a separate change which adds the overall progress to each line. e.g.:
```
1/5 Dumping 'my/package-1.2.0.0'
2/5 Dumping 'my/package-2.1.0.0'
3/5 Dumping 'my/package-2.0.0.0'
4/5 Dumping 'my/package-1.0.0.0'
5/5 Dumping 'my/package-1.1.0.0'
```

naderman:
Think that's something we can live with, and yes would be nice if you could send another PR with the numbering :-)

PR #229, opened 2015-06-30
Fixed typo in build command

gammamatrix:
The build command specifies using composer.json as the build file. This should be something like config.json.

See the documentation on: https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md
- This is the correct info: `php bin/satis build <configuration file> <build dir>`

The build command step on README.md and handling-private-packages-with-satis.md will be the same with this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Usage
 
 - Download [Composer](https://getcomposer.org/download/): `curl -sS https://getcomposer.org/installer | php`
 - Install satis: `php composer.phar create-project composer/satis --stability=dev --keep-vcs`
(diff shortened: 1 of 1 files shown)


gammamatrix:
There are a few spots where config.json is mentioned on handling-private-packages-with-satis.md instead of satis.json.

For consistency, should they be changed to satis.json since the documentation says that is the default?

naderman:
Yeah we should use satis.json consistently as config.json is also what you have in your composer home dir.

gammamatrix:
@naderman I created a pull request for changing `config.json` to `satis.json`: https://github.com/composer/composer/pull/4205

PR #232, opened 2015-07-14
forwardport PR#110

Tyrael:
this PR is basically the same as #110 but it should apply to the current master without any conflicts.

### DIFF
--- a/src/Composer/Satis/Command/BuildCommand.php
+++ b/src/Composer/Satis/Command/BuildCommand.php
@@ -146,36 +146,22 @@ EOT
             throw new \InvalidArgumentException('The output dir must be specified as second argument or be configured inside '.$input->getArgument('file'));
         }
 
-        if (($singleRepositoryUrl = $input->getOption('repository-url')) !== null
-            && FALSE !== ($otherPackagesCache = @file_get_contents($outputDir.'/packages.cache')))
-        {
-            $otherPackages = unserialize($otherPackagesCache);
-
-            // find repository configuration
+        if (($singleRepositoryUrl = $input->getOption('repository-url')) !== null) {
             $singleRepositoryConfig = null;
             foreach ($config['repositories'] as $r) {
                 if ($r['url'] == $singleRepositoryUrl) {
                     $singleRepositoryConfig = $r;
                     break;
-                }  
+                }
             }
             if ($singleRepositoryConfig === null) {
                 throw new \InvalidArgumentException('Requested repository not found in configuration: '.$singleRepositoryUrl);
             }
-
-            // use only selected repository
             $config['repositories'] = array($singleRepositoryConfig);
-
-            $composer = $this->getApplication()->getComposer(true, $config);
-            $packages = $this->selectPackages($composer, $output, $verbose, $requireAll, $requireDependencies, $requireDevDependencies, $minimumStability, $skipErrors, $packagesFilter);
-
-            // merge with cached data
-            $packages = array_merge($otherPackages, $packages);
-            ksort($packages, SORT_STRING);
-        } else {
-            $composer = $this->getApplication()->getComposer(true, $config);
-            $packages = $this->selectPackages($composer, $output, $verbose, $requireAll, $requireDependencies, $requireDevDependencies, $minimumStability, $skipErrors, $packagesFilter);
         }
+
+        $composer = $this->getApplication()->getComposer(true, $config);
+        $packages = $this->selectPackages($composer, $output, $verbose, $requireAll, $requireDependencies, $requireDevDependencies, $minimumStability, $skipErrors, $packagesFilter);
 
         if ($htmlView = !$input->getOption('no-html-output')) {
             $htmlView = !isset($config['output-html']) || $config['output-html'];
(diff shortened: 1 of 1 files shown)


samifruit514:
I don't get why this change is not merged yet. It is very useful when we push new tags into a git repo.

mbrodala:
@Seldaek What's your opinion on this one?

Given the fact that webhooks from Github/Bitbucket only contain Git repository URLs and not Composer package names, this change would be very useful to trigger updates for single packages.

naderman:
Finally got around to look at this. So first thing that comes to mind is: Why doesn't this use the merging of packages as the package filter does, see https://github.com/Tyrael/satis/blob/partial_update/src/Composer/Satis/Command/BuildCommand.php#L195 - is the cache file really needed?

Tyrael:
#110 was created before the packages filter was added (originally with 4e9780a980f3270eda98ba9a5797cce9463974f0 then later renamed)

naderman:
@Tyrael I realize the history, but would be nice to update this now then if possible?

Tyrael:
sure, and I will update the PR, I was just answering your question why was this not done that way in the first place :)

Tyrael:
@naderman any chance you had the time to review this?

src/Composer/Satis/Command/BuildCommand.php
```diff
@@ -145,6 +146,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
             throw new \InvalidArgumentException('The output dir must be specified as second argument or be configured inside '.$input->getArgument('file'));
         }
 
+        if (($singleRepositoryUrl = $input->getOption('repository-url')) !== null) {
+            $singleRepositoryConfig = null;
+            foreach ($config['repositories'] as $r) {
+                if ($r['url'] == $singleRepositoryUrl) {
```
stof: should use strict comparison

src/Composer/Satis/Command/BuildCommand.php
```diff
@@ -145,6 +146,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
             throw new \InvalidArgumentException('The output dir must be specified as second argument or be configured inside '.$input->getArgument('file'));
         }
 
+        if (($singleRepositoryUrl = $input->getOption('repository-url')) !== null) {
+            $singleRepositoryConfig = null;
+            foreach ($config['repositories'] as $r) {
```
stof: you cannot assume that this array key exists. (nor that ``url`` exists in it) as ``$config`` is just the json-decoded content of the config file, and so you have no idea about what it contains or what is not in it.

Tyrael:
while going through the comments from @stof I've just realized that @naderman suggestion in https://github.com/composer/satis/pull/232#issuecomment-138104593 was probably that we should instead extend selectPackages() and loadDumpedPackages() with a $repositoriesFilter similarly how it was done for $packagesFilter.
@naderman if that's what you wanted just let me now and I can do that.

realshadow:
How does it look? I could really use this, I have patched satis for now

alcohol:
It needs to be rebased unfortunately.

realshadow:
@Tyrael Can you please try to move this forward if you have time?

Or I can try it if you are busy.

Tyrael:
originally I just forwardported the quick and dirty solution but then I realized/was suggested a better approach but then got swamped with work at my dayjob and couldn't finish it.
I'm not given up on this but I don't think I can work on it for like another 2-3 weeks.

realshadow:
Just for clarification...

I assume combining `--repository-url` and `package` in one command is reduntant (and currently breaks packages). Should they be handled as mutually exclusive? 

Wouldn't it be better/easier to reuse `package` for repository URL as well and then based on argument type (repository URL, package name) get correct repository? Even though that might be a bit confusing as far as arguments go.

realshadow:
I looked at the latest code and it moved miles ahead...

After a bit of searching I found that this could be easily solved with just changing `findPackages` in `Composer\Repository\ArrayRepository` to include comparison of `$package->getSourceUrl()` (or create a separate method for that) and then it could be used together with package filter, e.g.

```
php bin\satis build satis.json public psr/log https://github.com/realshadow/serializers.git
```

Although I am not sure if it should go there. Any thoughts @alcohol, @Tyrael ?

alcohol:
@realshadow you mean you want to introduce changes to Composer for a feature in Satis? That is unlikely to happen unless you have a really good reason. Also I kind of lost track of what this PR was trying to accomplish since it's a port of a port (and I don't use Satis myself).

realshadow:
@alcohol Not really. Just saying :) 

It was trying to accomplish building of a single/multiple repositories with repository url instead of package name. The original proposal was for `--repository-url` argument. @Tyrael  then wanted to do  like suggested in https://github.com/composer/satis/pull/232#issuecomment-142918732, but since then    satis moved miles ahead and both mentioned methods do not exist anymore, instead the filtering is handled by composer itself via `findPackages` (thus my comment above).

Although on second thought, the original code, with slight alterations should be enough to cover this use case...

alcohol:
The code was mostly broken into smaller manageable pieces. It should still exist to some extend.

realshadow:
I have created a new PR https://github.com/composer/satis/pull/266 which uses `PackageSelection` for filtering.

JamesRezo:
I guess it can be closed ;)

PR #240, opened 2015-08-05
Fix type hint in BuildCommand

JeroenDeDauw:


alcohol:
These are two different interfaces. How do you know your change will not break anything?

JeroenDeDauw:
I verified that the only thing that is getting passed to this method is a `RootPackageInterface`. It'd be possible to use `CompletePackageInterface` as well, however `PackageInterface` does not contain all the methods that the code is actually invoking on the parameter, so using that as type hint is wrong.

alcohol:
This PR is in a conflicting state. Do you wish to update it, or can it be closed?

Issue #66, opened 2013-04-18
Specify notify-batch url from satis.json


cmodijk:
Is there a way to add the notify-batch from the satis.json file to be included in the packages.json file?

cmodijk:
We moved away from satis and because this ticket is 2 years old i'm closing this.

anroots:
Nevertheless, this feature would still be useful for smaller distributors who host a Satis server and want to know the download stats for their packages.

PR #246, opened 2015-09-10
Support "notify-batch" key in the JSON config file

anroots:
The value of this key - when present in config - will be included in the generated packages.json file.

This allows Satis repositories to track package downloads as detailed in https://getcomposer.org/doc/05-repositories.md#notify-batch


anroots:
This should solve #66.

JamesRezo:
This can be closed ;)

PR #253, opened 2015-10-13
Wip builder classes

JamesRezo:
This is an attempt to put the business logic of Satis into specific objects (Builders). The goal is to extend and/or reuse the business logic in other tools and maybe to simplify tests production.

### DIFF
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -21,7 +21,7 @@ use Composer\IO\ConsoleIO;
  *
  * @author James Hautot <james@rezo.net>
  */
-class DownloadsBuilder extends Builder implements BuilderInterface
+class ArchiveBuilder extends Builder implements BuilderInterface
 {
     /** @var InputInterface $input The input Interface. */
     private $input;
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -64,7 +64,7 @@ class ArchiveBuilder extends Builder implements BuilderInterface
         shuffle($packages);
         /* @var \Composer\Package\CompletePackage $package */
         foreach ($packages as $package) {
-            if ($helper->skipDump($package)) {
+            if ($helper->isSkippable($package)) {
                 continue;
             }
 
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -68,7 +68,7 @@ class ArchiveBuilder extends Builder implements BuilderInterface
                 continue;
             }
 
-            $this->output->writeln(sprintf("<info>Dumping '%s'.</info>", $name));
+            $this->output->writeln(sprintf("<info>Dumping '%s'.</info>", $package->getName()));
 
             try {
                 if ('pear-library' === $package->getType()) {
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -15,6 +15,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Helper\HelperSet;
 use Composer\Factory;
 use Composer\IO\ConsoleIO;
+use Composer\Package\PackageInterface;
 
 /**
  * Builds the archives of the repository.
@@ -36,19 +37,14 @@ class ArchiveBuilder extends Builder implements BuilderInterface
      */
     public function dump(array $packages)
     {
-        if (isset($this->config['archive']['absolute-directory'])) {
-            $directory = $this->config['archive']['absolute-directory'];
-        } else {
-            $directory = sprintf('%s/%s', $this->outputDir, $this->config['archive']['directory']);
-        }
+        $helper = new ArchiveBuilderHelper($this->output, $this->config['archive']);
+
+        $directory = $helper->getDirectory($this->outputDir);
 
         $this->output->writeln(sprintf("<info>Creating local downloads in '%s'</info>", $directory));
 
         $format = isset($this->config['archive']['format']) ? $this->config['archive']['format'] : 'zip';
         $endpoint = isset($this->config['archive']['prefix-url']) ? $this->config['archive']['prefix-url'] : $this->config['homepage'];
-        $skipDev = isset($this->config['archive']['skip-dev']) ? (bool) $this->config['archive']['skip-dev'] : false;
-        $whitelist = isset($this->config['archive']['whitelist']) ? (array) $this->config['archive']['whitelist'] : array();
-        $blacklist = isset($this->config['archive']['blacklist']) ? (array) $this->config['archive']['blacklist'] : array();
 
         $includeArchiveChecksum = isset($this->config['archive']['checksum']) ? (bool) $this->config['archive']['checksum'] : true;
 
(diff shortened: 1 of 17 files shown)


src/Composer/Satis/Builder/PackagesBuilder.php
```diff
@@ -0,0 +1,312 @@
...
+use Composer\Package\BasePackage;
+use Composer\Repository\PlatformRepository;
+use Composer\Package\Loader\ArrayLoader;
+use Composer\Package\Dumper\ArrayDumper;
+use Composer\DependencyResolver\Pool;
+use Composer\Json\JsonFile;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Builds the JSON stuff.
```
alcohol: 'stuff' does not really qualify as a description..

src/Composer/Satis/Builder/DownloadsBuilder.php
```diff
@@ -0,0 +1,159 @@
...
+use Symfony\Component\Console\Helper\HelperSet;
+use Composer\Factory;
+use Composer\IO\ConsoleIO;
+
+/**
+ * Builds the archives of the repository.
+ *
+ * @author James Hautot <james@rezo.net>
+ */
+class DownloadsBuilder extends Builder implements BuilderInterface
```
alcohol: Downloads is rather generic. Can you find something more specific to the purpose of this class?

src/Composer/Satis/Builder/Builder.php
```diff
@@ -0,0 +1,50 @@
...
+    /** @var OutputInterface $output The output Interface. */
+    protected $output;
+
+    /** @var string $outputDir The directory where to build. */
+    protected $outputDir;
+
+    /** @var array $config The parameters from ./satis.json. */
+    protected $config;
+
+    /** @var bool $skipErrors Escapes Exceptions if true. */
```
alcohol: Skips, not escapes.

src/Composer/Satis/Builder/Builder.php
```diff
@@ -0,0 +1,50 @@
...
+    protected $skipErrors;
+
+    /**
+     * Base Constructor.
+     *
+     * @param OutputInterface $output     The output Interface
+     * @param string          $outputDir  The directory where to build
+     * @param array           $config     The parameters from ./satis.json
+     * @param bool            $skipErrors Escapes Exceptions if true
```
alcohol: Same as previous comment.

src/Composer/Satis/Builder/DownloadsBuilder.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /** @var HelperSet $helperSet A HelperSet instance. */
+    private $helperSet;
+
+    /**
+     * Builds the archives of the repository.
+     *
+     * @param array $packages List of packages to dump
+     */
+    public function dump(array $packages)
```
alcohol: This method still contains a **lot** of logic. It should just be dumping. Try to extract the white/black-listing and such into helper methods or even other classes possibly,

src/Composer/Satis/Builder/DownloadsBuilder.php
JamesRezo: it cames from the original function name **dumpDownloads**. **ArchiveBuilder**, reffering to the config property, is perhaps a better name ?

src/Composer/Satis/Builder/PackagesBuilder.php
JamesRezo: Right. "Builds the JSON files" is better

src/Composer/Satis/Builder/DownloadsBuilder.php
JamesRezo: I agree. It is also true for PackagesBuilder->select() 
I can work on both ...

alcohol: If you're willing to put in the time and work, that would be appreciated.

src/Composer/Satis/Builder/DownloadsBuilder.php
alcohol: Was going to be my suggestion too yeah.

JamesRezo:
Commit 84e99d61ad2a8ac5a0a4279fe1ebfbdf435139c2 (and the next...) provides a Helper Class for the renamed **ArchiveBuilder** class. And unit tests for this new object ;)

I am wondering now if it lowers the logic enough or not. For example, if it were useful to extract the pear-library test in a private doDump()-like method....

I also don't understand the presence of code after the effective dump. There are some setDistxxx methods applied to $package and, as far as i know, with no chance to throw an exception...

JamesRezo:
Commit 2f4ec99745fd86232375f2d56bf73c1100acb9d2 disptach the logic of Packages selection with a few helper methods.
I can still work for other improvements until saturday. After that, I'll be away for three weeks ;)

src/Composer/Satis/Builder/ArchiveBuilderHelper.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Tells if a package has to be dumped.
+     *
+     * @param  PackageInterface $package The package to be dumped
+     *
+     * @return bool false if the package has to be dumped.
+     */
+    public function skipDump(PackageInterface $package)
```
alcohol: I would sooner opt for a name such as `isSkippable`, `isDumpable` (negating/inverse the result).

The function name should say something about the subject, `skipDump` relates more to the operator (the helper).

src/Composer/Satis/Builder/ArchiveBuilder.php
```diff
@@ -0,0 +1,137 @@
...
+
+        /* @var \Composer\Downloader\DownloadManager $downloadManager */
+        $downloadManager = $factory->createDownloadManager($io, $composerConfig);
+
+        /* @var \Composer\Package\Archiver\ArchiveManager $archiveManager */
+        $archiveManager = $factory->createArchiveManager($composerConfig, $downloadManager);
+
+        $archiveManager->setOverwriteFiles(false);
+
+        shuffle($packages);
```
alcohol: Is there any reason for shuffling the packages?

JamesRezo: The purpose of this PR is just refactoring, not changing the behavior, so I just copied the code :)

shuffling was proposed by @joshdifabio and merged by @naderman in #225

alcohol: Ah, interesting.

src/Composer/Satis/Builder/ArchiveBuilderHelper.php
JamesRezo: isSkippable() is a quick fix, i'd prefer this one

JamesRezo:
Well... I can't do more for now. Archive dump tests are a bit difficult to do in a short time. So are Package Selection tests.

@alcohol : Can we consider, nevertheless, that this proposal is sufficient?

alcohol:
Well, I can YOLO it. How solid do you feel about your PR? Have you ran it yourself?

JamesRezo:
I've ran it on two personnal projects and it seemed ok for me.
Truth is there are no tests in satis today to check for regressions... and as I said, i'm away for 3 weeks.
So... How do you feel ? :)

alcohol:
@naderman how do you feel about this PR?

* [ ] happy
* [ ] sad
* [ ] indifferent

naderman:
Thanks!

Iruwen:
Stopped working for me after this merge, I'm now getting:
"PHP Fatal error:  Class 'Composer\Satis\Builder\Filesystem' not found in /var/www/satis/src/Composer/Satis/Builder/ArchiveBuilder.php on line 76"

alcohol:
Should be fixed by b7fa484

Iruwen:
That was fast. Works.

PR #258, opened 2015-10-28
customer downloader for archiving packages with unknown downloader type

kschu91:
We´re using satis archive funcionality to mirror all our dependencies. Since we have a TYPO3 project, a lot of our dependecies are downloaded from https://composer.typo3.org/ which provide packages in "t3x" format.
Satis does not support any confiugation to specify custom downloader types. Thus, we have implemented it.

Enable custom downloader with following configuration within the "archive" property.
```json
"downloader": [
        {
            "type": "t3x",
            "file": "build/downloader/T3xDownloader.php",
            "class": "\\TYPO3\\CMS\\Composer\\Installer\\Downloader\\T3xDownloader"
        }
]
```

would be nice to have it in the master :smile: 

Btw. Thanks for this awesome tool :wink:

### DIFF
--- a/src/Composer/Satis/Command/BuildCommand.php
+++ b/src/Composer/Satis/Command/BuildCommand.php
@@ -650,7 +650,7 @@ EOT
                 );
             }
 
-            $file = $customDownloader['file'];
+            $file = getcwd() . DIRECTORY_SEPARATOR . $customDownloader['file'];
             $class = $customDownloader['class'];
             $type = $customDownloader['type'];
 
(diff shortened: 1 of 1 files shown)


alcohol:
If you could rebase your PR and fix some minor PSR2 issues, I'm ok with merging this.

JamesRezo:
I could be wrong, but I think it could be a better implementation to use at this point the $composer instantiated in the `exectute()` method of BuildCommand. If my understanding is correct, $composer loads composer plugins and has InstallationManager and DownloadManager instances set. If true, doing this rather than implement a specific autoloader could satisfy the case.

JamesRezo:
@kschu91 could you test #265 please ?

PR #264, opened 2015-11-13
Splitting PackagesBuilder with a new PackageSelection Object

JamesRezo:
following up the goal of #253 :
+ it splits the business logics of package selection and packages build
+ Unit Tests


alcohol:
If you could add a `.travis.yml` that would be great :-)

Initially you can look at the one [composer/semver](/composer/semver) or [composer/spdx-license](/composer/spdx-licenses) use.

JamesRezo:
Done ! :)

.travis.yml
```diff
@@ -0,0 +1,34 @@
+language: php
+
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache
```
stof: I suggest caching only the ``/files`` subfolder. See https://github.com/puli/manager/pull/9#discussion_r40499480 for the explanation

alcohol:
Thanks :)

@stof you recommend I also modify this caching path on the other composer repos?

PR #265, opened 2015-11-19
Allow ArchiveBuilder to use a composer plugin

JamesRezo:
This is an alternative implementation of #258 that allow the ArchiveBuilder to use a composer plugin that sets a custom Downloader. It implies the plugin to be globally installed with composer (or installed where `COMPOSER_HOME` points to).


alcohol:
Thanks

PR #220, opened 2015-05-19
Add an abandoned config key to mark packages as so

atimmer:
Fixes #183, #213.

I chose the straightforward approach, just a new config option with key value pairs to define abandonment of packages. Let me know if this is the correct approach.

PR #268, opened 2015-11-22
Feature/abandoning/rebase

JamesRezo:
This is a rebase of #220 to add abonned marks from packages and/or configuration file
This PR provides also some Unit tests of the feature


alcohol:
You have committed a patch file. I think that was unintended?

src/Composer/Satis/PackageSelection/PackageSelection.php
```diff
@@ -284,6 +290,19 @@ public function load()
     }
 
     /**
+     * Mark packages as abandoned by Configuration file
+     */
+    private function setAbandoned()
```
alcohol: I'm not going to ask you to change this, but I do have a suggestion for future PR's.

Using methods on the class like this internally in the function makes them a teensy bit uglier to test. It is always better to pass in dependencies of methods. It also makes it more clear from the function signature what it works with at that point then. Cause right now, you have to read through the whole function body to determine what it could possibly be doing/changing.

JamesRezo: I agree.

alcohol:
Not exactly what I suggested btw, but thanks :)

SvenRtbg:
Does this come with documentation here: https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md ?

src/Composer/Satis/PackageSelection/PackageSelection.php
```diff
@@ -284,6 +290,19 @@ public function load()
...
 
     /**
+     * Marks selected packages as abandoned by Configuration file
+     */
+    private function setSelectedAsAbandoned()
+    {
+        foreach ($this->selected as $name => $package) {
+            if (array_key_exists($package->getName(), $this->abandoned)) {
+                $package->setAbandoned($this->abandoned[$package->getName()]);
+                $this->selected[$package->getUniqueName()] = $package;
```
stof: this line looks useless to me

JamesRezo: You're right. This line can be safely removed.
Does it need a PR ?

JamesRezo:
@SvenRtbg Indeed, the documentation needs to be updated. 
As I'm working on a satis file validation schema, I was planning to PR this schema before such an update. But feel free to start the job ! :)
Considering documentation, I'm wondering if other missing properties (whitelist/blacklist), have to be mentioned.
Moreover, as the schema evolved, it think it might be useful to release an alpha2...

src/Composer/Satis/PackageSelection/PackageSelection.php
alcohol: A PR doesn't have to be big. Even 1 line changes are welcome.

alcohol:
@JamesRezo if you want to work on validation, look at how Composer does it (if you have not done so already). :)

PR #271, opened 2015-11-25
satis.json schema

JamesRezo:
This is a proposal for a satis file validation schema.
It can be tested with https://github.com/justinrainbow/json-schema : 
```shell
validate-json /path/to/satis.json res/satis-schema.json
```

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -21,7 +21,11 @@
         "require": {
             "type": "object",
             "description": "This is a hash of package name (keys) and version constraints (values) that are required to run this package.",
-            "additionalProperties": true
+            "minProperties": 1,
+            "additionalProperties": {
+                "type": "string",
+                "description": "A valid version constraint"
+            }
         },
         "archive": {
             "type": "object",
@@ -75,12 +79,15 @@
         "minimum-stability": {
             "type": ["string"],
             "description": "The minimum stability the packages must have to be install-able. Possible values are: dev, alpha, beta, RC, stable.",
-            "pattern": "^dev|alpha|beta|rc|RC|stable$"
+            "enum": [ "dev", "alpha", "beta", "rc", "RC", "stable" ]
         },
         "abandoned": {
             "type": "object",
             "description": "List of packages marked as abandoned for this repository, the mark can be boolean or a package name/URL pointing to a recommended alternative.",
-            "additionalProperties": true
+            "additionalProperties": {
+                "type": ["boolean", "string"],
+                "description": "A valid Package name"
+            }
         },
         "require-all": {
             "type": "boolean",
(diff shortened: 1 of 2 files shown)


res/satis-schema.json
```diff
@@ -0,0 +1,115 @@
...
+        },
+        "repositories": {
+            "type": ["object", "array"],
+            "description": "A set of additional repositories where packages can be found.",
+            "additionalProperties": true
+        },
+        "minimum-stability": {
+            "type": ["string"],
+            "description": "The minimum stability the packages must have to be install-able. Possible values are: dev, alpha, beta, RC, stable.",
+            "pattern": "^dev|alpha|beta|rc|RC|stable$"
```
stof: using ``enum`` would be simpler

alcohol:
It would probably be a good idea to validate the schema on every run. Not necessarily fail the run, but at least print out errors/warnings to notify the user.

res/satis-schema.json
```diff
@@ -0,0 +1,115 @@
...
+        },
+        "homepage": {
+            "type": "string",
+            "description": "Homepage URL for the project.",
+            "format": "uri"
+        },
+        "require": {
+            "type": "object",
+            "description": "This is a hash of package name (keys) and version constraints (values) that are required to run this package.",
+            "additionalProperties": true
```
stof: rather than ``true``, this should use a schema object here, to force all additional properties to contain strings

res/satis-schema.json
```diff
@@ -0,0 +1,115 @@
...
+        },
+        "minimum-stability": {
+            "type": ["string"],
+            "description": "The minimum stability the packages must have to be install-able. Possible values are: dev, alpha, beta, RC, stable.",
+            "pattern": "^dev|alpha|beta|rc|RC|stable$"
+        },
+        "abandoned": {
+            "type": "object",
+            "description": "List of packages marked as abandoned for this repository, the mark can be boolean or a package name/URL pointing to a recommended alternative.",
+            "additionalProperties": true
```
stof: same here about using a schema

res/satis-schema.json
```diff
@@ -0,0 +1,115 @@
...
+            "type": "boolean",
+            "description": "If true, build a static web page."
+        },
+        "twig-template": {
+            "type": "string",
+            "description": "Path to a template for the static web page."
+        },
+        "extra": {
+            "type": ["object", "array"],
+            "description": "Arbitrary extra data that can be used by plugins, for example, package of type composer-plugin may have a 'class' key defining an installer class name.",
```
stof: the description is wrong. You are talking about the package schema here, not about the satis schema

stof:
You are missing the ``config`` property (which is the same than the property of the composer package schema)

JamesRezo:
@stof, i completely agree about `require`, that could be the same for `repositories`. But it's to Composer to support sub-schema I guess. Is it possible to have more than one validation file in Composer ?
I can of course add a `config` property and maybe something about authentication... but again, it's about to Composer to support such sub-schemas.
Also, `minimum-stability` is a copy of the Composer property.

What I mean is that it would be preferable Composer manages its definitions and then Satis could use it. How do you feel about this ?

I have no idea how to set `abandoned` specification here. It has to match
```json
"abandoned" : {
    "vendor/name": true,
    "vendor/othername": "othervendor/othername",
   ...
}
```
It's clearly an object, but how to specify its properties ?

@alcohol. Yep, it's the idea I have in mind. But for now, il would be a lot a duplicate code between Composer and Satis.

stof:
@JamesRezo ``require`` values are just strings, same for ``abandonned`` (well, string or boolean in this case).

stof:
> I can of course add a config property

if you don't add it at all, you forbid defining it in the satis config, which is wrong.

JamesRezo:
Sorry for the type mismatch :-)

res/satis-schema.json
```diff
@@ -0,0 +1,249 @@
...
+                    "type": "string",
+                    "description": "The default archiving format when not provided on cli, defaults to \"tar\"."
+                },
+                "archive-dir": {
+                    "type": "string",
+                    "description": "The default archive path when not provided on cli, defaults to \".\"."
+                }
+            }
+        },
+        "extra": {
```
stof: ``extra`` does not make sense in the Satis config

res/satis-schema.json
```diff
@@ -0,0 +1,249 @@
...
+                },
+                "github-expose-hostname": {
+                    "type": "boolean",
+                    "description": "Defaults to true. If set to false, the OAuth tokens created to access the github API will have a date instead of the machine hostname."
+                },
+                "archive-format": {
+                    "type": "string",
+                    "description": "The default archiving format when not provided on cli, defaults to \"tar\"."
+                },
+                "archive-dir": {
```
stof: these 2 should be removed, as the configuration is done through ``archive`` instead

src/Composer/Satis/Command/BuildCommand.php
```diff
@@ -119,6 +119,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $config = $file->read();
         }
 
+        try {
+            $this->check($configFile);
+        } catch (\Exception $e) {
+            if (!$skipErrors) {
+                throw $e;
+            }
+            $output->writeln(sprintf("<error>Skipping Exception '%s'.</error>", $e->getMessage()));
```
alcohol: Better to use warning instead of error I think, since `$skipErrors` basically downgrades them.

JamesRezo:
Is there something blocking yet ?

alcohol:
No, I'll merge it, and if anything pops up I expect a new PR :-)

glensc:
looks like this change made "homepage" key required, and it was not easy to figure out WHY the validation failed. i had pretty much test each option to figure out what made validation to fail...

```
$ satis build libs.json  -vvv

  [Composer\Json\JsonValidationException]                       
  The json config file does not match the expected JSON schema  
                                                                
Exception trace:
 () at /usr/share/php/Composer/Satis/Command/BuildCommand.php:269
 Composer\Satis\Command\BuildCommand->check() at /usr/share/php/Composer/Satis/Command/BuildCommand.php:128
 Composer\Satis\Command\BuildCommand->execute() at /usr/share/php/Symfony/Component/Console/Command/Command.php:256
 Symfony\Component\Console\Command\Command->run() at /usr/share/php/Symfony/Component/Console/Application.php:838
 Symfony\Component\Console\Application->doRunCommand() at /usr/share/php/Symfony/Component/Console/Application.php:189
 Symfony\Component\Console\Application->doRun() at /usr/share/php/Composer/Satis/Console/Application.php:52
 Composer\Satis\Console\Application->doRun() at /usr/share/php/Symfony/Component/Console/Application.php:120
 Symfony\Component\Console\Application->run() at /usr/bin/satis:13


build [--repository-url [REPOSITORY-URL]] [--no-html-output] [--skip-errors] [--] [<file>] [<output-dir>] [<packages>]...
$ cat libs.json
{
        "name": "libs",
        "description": "Composer Libraries Repository",
        "require-all": true,
        "repositories": [
        ]
}
```

glensc:
also, "name" seems to be required now, i had repo without name working fine before (generated without html output)

stof:
Well, the BuildCommand should display validation errors properly (the JsonValidationException contains some details)

PR #272, opened 2015-11-25
fix dumping all packages with --repository-filter

mathroc:
don't update them but still dump the existing version of all packages when `--repository-url` is used


alcohol:
Thanks

PR #275, opened 2015-11-28
notify-batch key, rebase #246

JamesRezo:
see #66 and #246 + Unit tests

### DIFF
--- a/tests/Composer/Test/Satis/PackagesBuilderDumpTest.php
+++ b/tests/Composer/Test/Satis/PackagesBuilderDumpTest.php
@@ -31,16 +31,7 @@ class PackagesBuilderDumpTest extends \PHPUnit_Framework_TestCase
     {
         $this->package = new Package('vendor/name', '1.0.0.0', '1.0');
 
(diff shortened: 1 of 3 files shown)


tests/Composer/Test/Satis/PackagesBuilderDumpTest.php
```diff
@@ -23,6 +23,26 @@
...
+    protected $root;
+
+    protected function setUp()
+    {
+        $this->package = new Package('vendor/name', '1.0.0.0', '1.0');
+
+        $this->root = $this->setFileSystem();
+    }
+
+    protected function setFileSystem()
```
alcohol: It's better to make this method return a `fileSystem()`. Using side-effects (e.g. changing state instead of returning something) is generally a bad idea.

JamesRezo: Well, ok about side-effect/change state, but I didn't find a way to change the code that way. Then, i simplified removing the method...

alcohol:
Cheers

PR #278, opened 2015-12-01
closes #277 (same idea, different implementation)

alcohol:


src/Composer/Satis/PackageSelection/PackageSelection.php
```diff
@@ -425,4 +422,30 @@ private function getRequired(PackageInterface $package)
 
         return $required;
     }
+
+    /**
+     * Filter given repositories.
+     *
+     * @param array $repositories
+     *
+     * @return array
```
stof: I would document it as ``RepositoryInterface[]`` instead

alcohol: Ah, sure can do that.

Issue #282, opened 2015-12-11
satis.phar is fat


JamesRezo:
Since unit tests are developped alongside satis, the Compiler builds a very big satis.phar and the needs on max open files is growing up.
This issue could be easily resolved by adding
```php
->exclude(array(
    'phpunit', 'mikey179',
    'phpdocumentor', 'sebastian', 'phpspec', 'doctrine',
    'test', 'tests', 'Tests'
))
```
in the Finder.

As nobody complains with this issue :), we could also replace that piece of code by a `box.json` file in order to use [box project](https://github.com/box-project/box2)

Whatever the choice, we also could add some documentation on that feature.

alcohol:
I quite like Box and would personally vote in favour of that approach.

/cc @Seldaek @naderman any objections/feedback?

Seldaek:
Nope. If it does the job it's fine by me.

PR #283, opened 2015-12-12
Replace self-made phar compiler by box

JamesRezo:
As discussed in #282 
before to run `box` on satis
```bash
ulimit -n 1024
```


alcohol:
I wouldn't remove the compiler just yet in the same PR. Maybe trigger a `E_USER_DEPRECATED` error instead from within it. It might very well be some people run it as part of some automated update + build process.

alcohol:
Any particular reason for the `ulimit -n 1024`? I've never had to do that myself.

JamesRezo:
On OS X, open max files is set to 256

alcohol:
I am on OSX. Oh well, I'll have to try it with this PR :)
Will give it a shot tomorrow, thanks.

JamesRezo:
OK for maintaining the compiler and box. So, I guess I have to do another PR (I don't know how to re-add removed files in a commit saving history)

alcohol:
I can do that for you, when I merge it, don't worry about it.

JamesRezo:
OK, thanks. Don't forget to add the `->exclude(...)` clause in the compiller then. ;)

alcohol:
Thanks

box.json
```diff
@@ -0,0 +1,29 @@
...
+    "alias": "satis.phar",
+    "main": "bin/satis",
+    "output": "satis.phar",
+    "stub": true,
+    "directories": [ "src", "views" ],
+    "files": [
+        "LICENSE"
+    ],
+    "finder": [{
+        "name": "*.php",
```
stof: what about other files which may be needed (for instance the exe file necessary to ask hidden questions on windows ?)

alcohol: The `Compiler` class does not include those either. So I'm guessing they're either never necessary or people rarely/never use satis as a Phar.

JamesRezo: Actually, Satis isn't an interactive application, as a Phar or not.
The only use case for interactivity could be when asking for username/password when getting a _remote_ `satis.json` file. But it doesn't happen because the io is a NullIO Object, always (AFAIK).

Issue #236, opened 2015-07-21
The directory of downloaded packages is more great than the packages published


carlososiel:
The directory of downloaded packages is more great than the packages published, this happen because the dev-master packages is downloaded again if exist a new commit in the repository, but it keeps previous versions of package though not are published. Exist any way to clean these obsolete packages? The system can be delete these packages automatically with the features that exist today? If not exist this feature can be included?

stof:
what do you mean with ``directory of downloaded packages`` ?

carlososiel:
The directory is web/dist, where the packages are stored in .zip format

stof:
Well, this folder is not about packages downloaded by Satis but generated by Satis to be downloadable.
The reason why Satis does not delete old ones is to keep them available for projects having a lock file referencing one of them (i.e. they have a dev version in their lock file, but new commits have been done in the meantime and Satis is regenerated).
There is no command to clean old dist files currently, but this could be implemented (with a warning to make people aware of the drawback)

joshdifabio:
@carlososiel Bear in mind that, although those dist files are no longer included in the `packages.json` file of your Satis instance, they may still be referenced in the `composer.lock` files of various projects which have been built using Composer. If you delete those files, `composer install` might fail for some of your projects until you run `composer update`.

aaukt:
There also is the [skip-dev](https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md#options-explained) option so things like dev-master are skipped when dumping, so it would not create a ne zip for every new commit.
This does of course not work when you depend on master being downloadable as a dist.

carlososiel:
@joshdifabio I understand the risks, and i proceed with much care, but is important for me include this feature in the system to keep only the packages used in my server.
@aaukt Is not a option for me, i use packages that depend of dev-master versions

aaukt:
@carlososiel dev-master is still available, it just links directly to the git repository (and composer would clone it)

curry684:
@carlososiel the 'feature' cannot be implemented, as by its very nature composer nor satis have any way of identifying which packages are 'obsolete', as they don't know what's out there in the world. Your feature request implies reverse knowledge which would honestly only apply in your specific use case. And for single-user features I would recommend just writing a script on your end instead that cleans them up, rather trivial given the versioned filenames.

mikebarlow:
Sorry to revive an old issue but it seemed more sensible then creating a new one.

[...]

PR #284, opened 2015-12-14
Clean folders

zyfran:
Hello. I want to propose changes to the code.

In our project we are faced with fact that when downloading all dependencies, the archives for versions which were removed from vcs are still present in output directory.

Also same problem present with "include/all$%hash%.json" files. These files are not removed after each build.

### DIFF
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -101,11 +101,19 @@ class ArchiveBuilder extends Builder implements BuilderInterface
                 if (!$this->skipErrors) {
                     throw $exception;
                 }
+
+                //If file exists then exclude it
+                $packageName = $archiveManager->getPackageFilename($package);
+                if ('pear-library' === $package->getType()) {
+                    $excludeFiles[] = $packageName.'.'. pathinfo($package->getDistUrl(), PATHINFO_EXTENSION);
+                } else {
+                    $excludeFiles[] = $packageName . '.' . $format;
+                }
                 $this->output->writeln(sprintf("<error>Skipping Exception '%s'.</error>", $exception->getMessage()));
             }
         }
 
-        $this->clean($directory, $excludeFiles);
+        $this->clean($directory, array_unique($excludeFiles));
     }
 
     /**
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -55,6 +55,7 @@ class ArchiveBuilder extends Builder implements BuilderInterface
         $archiveManager->setOverwriteFiles(false);
 
         shuffle($packages);
+        $excludeFiles = array();
         /* @var \Composer\Package\CompletePackage $package */
         foreach ($packages as $package) {
             if ($helper->isSkippable($package)) {
@@ -86,6 +87,7 @@ class ArchiveBuilder extends Builder implements BuilderInterface
                     $archiveFormat = $format;
                 }
                 $archive = basename($path);
+                $excludeFiles[] = $archive;
                 $distUrl = sprintf('%s/%s/%s', $endpoint, $this->config['archive']['directory'], $archive);
                 $package->setDistType($archiveFormat);
                 $package->setDistUrl($distUrl);
@@ -102,6 +104,8 @@ class ArchiveBuilder extends Builder implements BuilderInterface
                 $this->output->writeln(sprintf("<error>Skipping Exception '%s'.</error>", $exception->getMessage()));
             }
         }
+
+        $this->clean($directory, $excludeFiles);
     }
 
     /**
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -113,7 +113,7 @@ class ArchiveBuilder extends Builder implements BuilderInterface
             }
         }
 
-        $this->clean($directory, array_unique($excludeFiles));
+        $this->removeAllBut($directory, array_unique($excludeFiles));
     }
 
     /**
--- a/src/Composer/Satis/Builder/Builder.php
+++ b/src/Composer/Satis/Builder/Builder.php
@@ -53,11 +53,19 @@ class Builder
      *
      * @param string          $directory    Directory to clean
      * @param string|string[] $excludeFiles File list to exclude
+     * @param bool            $forceRemove  Forcibly remove files
      *
      * @return void
      */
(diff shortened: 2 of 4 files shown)


zyfran:
issue #236 can be fixed by my commits

zyfran:
When the archive was deleted and `composer.lock` exist
`composer install` output:

- Installing vendor/package (vX.Y.Z)  
  Downloading: 0%    Failed to download vendor/package from dist: The "http://localhost:8000/dist/vendor-package-vX.Y.Z-78ba13.zip" file could not be downloaded (HTTP/1.1 404 Not Found)  
  Now trying to download from source
- Installing vendor/package (vX.Y.Z)  
  Cloning a217e78a49f1eb2548bad1fd81c8b7fd84124b15

stof:
@zyfran keeping the old archives is necessary in some cases. For instance, if your Satis repo includes dev versions of packages, you don't want to remove archives for all older commits of the dev versions, because some projects may still rely on them in their lock file, while the new build will have generated the archive for the latest version of the branch (but which will be picked by the project only after their next ``composer update``)

src/Composer/Satis/Builder/Builder.php
```diff
@@ -47,4 +47,29 @@ public function __construct(OutputInterface $output, $outputDir, $config, $skipE
...
+
+    /**
+     * Remove from $directory all files excluding $excludeFiles
+     *
+     * @param string          $directory    Directory to clean
+     * @param string|string[] $excludeFiles File list to exclude
+     *
+     * @return void
+     */
+    protected function clean($directory, $excludeFiles)
```
stof: I don't thinks this has any reason to be in the base class, as only one of the child classes actually needs it

stof: ah no sorry

zyfran: @stof  PackagesBuilder and ArchiveBuilder extends Builder

src/Composer/Satis/Builder/ArchiveBuilder.php
```diff
@@ -99,9 +101,19 @@ public function dump(array $packages)
                 if (!$this->skipErrors) {
                     throw $exception;
                 }
+
+                //If file exists then exclude it
```
alcohol: Why would you do this in the `catch` of a `try`/`catch` block?

src/Composer/Satis/Builder/PackagesBuilder.php
```diff
@@ -83,6 +83,8 @@ private function dumpPackageIncludeJson(array $packages)
         rename($this->filenamePrefix, $filenameWithHash);
         $this->output->writeln("<info>wrote packages json $filenameWithHash</info>");
 
+        $this->clean($this->outputDir . '/include', "all\$$hash.json");
```
JamesRezo: the consequence here is to delete the file you've just created.

```
$this->filenamePrefix = $this->outputDir.'/include/all'; #at construction
$filenameWithHash = $this->filenamePrefix.'$'.$hash.'.json'; #at dump
```
Then, the clean() unlinks `$this->outputDir. '/include/' . "all\$$hash.json"`

This breaks the Packages Builder tests.

src/Composer/Satis/Builder/ArchiveBuilder.php
```diff
@@ -99,9 +101,19 @@ public function dump(array $packages)
...
+                $packageName = $archiveManager->getPackageFilename($package);
+                if ('pear-library' === $package->getType()) {
+                    $excludeFiles[] = $packageName.'.'. pathinfo($package->getDistUrl(), PATHINFO_EXTENSION);
+                } else {
+                    $excludeFiles[] = $packageName . '.' . $format;
+                }
```
JamesRezo: In both cases, this is the value of `$path`, isn't it ?

src/Composer/Satis/Builder/PackagesBuilder.php
JamesRezo: Oops, sorry. clean() method unlinks all but that file. :)

src/Composer/Satis/Builder/ArchiveBuilder.php
zyfran: @alcohol Because if you do not add the name of a file into `$excludeFiles`, it will be removed

src/Composer/Satis/Builder/ArchiveBuilder.php
zyfran: @JamesRezo `$path = $archiveManager->archive($package, $format, $directory);` can throw Exception and `$path` will be:
- null
- `$path = realpath($directory).'/'.$packageName.'.'.pathinfo($package->getDistUrl(), PATHINFO_EXTENSION);`
- value set in the previous iteration

src/Composer/Satis/Builder/PackagesBuilder.php
zyfran: `It's name is confusing` I agree )

JamesRezo:
Well, as a purge procedure could be a good idea, having a systematic cleaning is maybe a bit too tough. No rollback will be possible, even manual, for example. I'd prefer a Purge command. But for now, a config key in the configuration file, set to `false` by default, could fit the need ?

src/Composer/Satis/Builder/ArchiveBuilder.php
JamesRezo: So, what will happen if no exception is catched ?

zyfran:
Perhaps I can agree with the need to add a configuration key. When its value `true` archives will be deleted. But for JSON files I think it is not necessary to use this key because when rebuilding project they are no longer used.

src/Composer/Satis/Builder/ArchiveBuilder.php
zyfran: @JamesRezo if `$this->skipErrors===false` then `dump()` method will stop and no files will be deleted

JamesRezo:
`they are no longer used.` except for rollbacks, that are very rare, i admit.

zyfran:
Well, i think that there must be RollbackCommand and before rebuilding project all json files must be saved to some dump directory

JamesRezo:
Yes, if the need were common, why not. It can be done another way : saving by script before build, restoring if something went wrong... Not a serious deal.

zyfran:
- Rename `clean()` method to `removeAllBut()`
- Add `$forceRemove` param to `removeAllBut()` method
- Add `autoclean` config key with default value `false`

zyfran:
Add `autoclean` property to `satis-schema.json`

alcohol:
Is this still necessary now that the purge command has been merged in?

PR #292, opened 2016-01-06
add support configuring include filename

glensc:
this solves problem that includes/ dir grows with stale json files by allowing to configure fixed name:

```json
{
    "name": "...",
    "description": "...",
    "include-filename": "all.json",
    "repositories": [
```

```
$ satis/bin/satis build satis.json dist
Scanning packages
Initializing PEAR repository http://pear.php.net
Wrote packages json all.json
Writing packages.json
Writing web view
```

ps: side effect is that `hash_file` is no longer called twice ;)

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -105,6 +105,10 @@
             "type": "boolean",
             "description": "If true, resolve and add all Dev dependencies of each required package."
         },
+        "include-filename": {
+            "type": "string",
+            "description": "Specify filename instead of default include/all${SHA1_HASH}.json"
+        },
         "output-dir": {
             "type": "string",
             "description": "The directory where the static Repository is built."
--- a/src/Composer/Satis/Builder/PackagesBuilder.php
+++ b/src/Composer/Satis/Builder/PackagesBuilder.php
@@ -11,7 +11,6 @@
 
 namespace Composer\Satis\Builder;
 
-use Composer\Composer;
 use Composer\Json\JsonFile;
 use Composer\Package\Dumper\ArrayDumper;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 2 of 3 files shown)


glensc:
comments welcome on following topics:
-  should the option be moved under "config" key?
- is the "include-filename" good name choice here?

glensc:
never mind about moving to "config", the "config" is for [composer options](https://github.com/composer/satis/blob/acc1cf66dde8be98111902e4b8b3b3e84b43a426/res/satis-schema.json#L120-L122)!

glensc:
can someone review this? add comments?

glensc:
@alcohol can you take look?

alcohol:
Afaik the hash is used by Composer to validate the contents of the file though. Allowing users to override the name without specifying a placeholder for the sha1 hash might break said functionality?

glensc:
the hash is still preserved in json content, this change only affects include filename.

alcohol:
Yes but Composer takes the hash from the filename.

glensc:
no, composer takes filename that is defined in package.json

example: https://github.com/eventum/composer/blob/gh-pages/packages.json
```json
{
    "includes": {
        "all.json": {
            "sha1": "3f8449992f4151af865486a74c0f99c1679cf827"
        }
    }
}
```

thus it  reads https://github.com/eventum/composer/blob/gh-pages/all.json next and verifies `3f8449992f4151af865486a74c0f99c1679cf827` as hash

alcohol:
Oh, it seems we only do this for the provider files.

https://github.com/composer/composer/blob/cc75946ef23ffcc8f14122dc62306d747fd3ccd5/src/Composer/Repository/ComposerRepository.php#L544-L569

I guess it does not really serve a purpose in that case.

glensc:
thanks for the merge!

PR #295, opened 2016-01-18
Providers support

copitz:
This PR makes it possible to dump providers by setting providers: true in config.

Also it prunes old include and provider json files after package json dump in order to save disk space.

Running Unit Test included

### DIFF
--- a/src/Composer/Satis/Builder/PackagesBuilder.php
+++ b/src/Composer/Satis/Builder/PackagesBuilder.php
@@ -29,6 +29,8 @@ class PackagesBuilder extends Builder
     /** @var string included json filename template */
     private $includeFileName;
 
+    private $writtenIncludeJsons = array();
+
     /**
      * Dedicated Packages Constructor.
      *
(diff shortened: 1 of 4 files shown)


alcohol:
If you would be willing to rebase your PR, we can look at merging it.

copitz:
Done

aaukt:
From having build this myself i can tell you, that special care has to be taken of "replace" packages. The package that is replaced has to include the version information of its "replacee" in the json file.
As an example you can see how it is done in https://github.com/composer/composer/tree/master/doc/fixtures/repo-composer-with-providers

src/Composer/Satis/Builder/PackagesBuilder.php
```diff
@@ -126,29 +134,71 @@ private function pruneIncludeDirectories()
...
         );
     }
 
     /**
+     * Write to a file
+     *
+     * @param string $path
+     * @param string $contents
+     */
+    private function writeToFile($path, $contents)
```
glensc: i think this method needs explanation why it is here. what for the retries are there. it also looks like copy paste from some other library. imho should be noted so (via `@see URL`  perhaps)

src/Composer/Satis/Builder/PackagesBuilder.php
```diff
@@ -126,29 +134,71 @@ private function pruneIncludeDirectories()
...
+                );
+            }
+        }
+
+        $retries = 3;
+        while ($retries--) {
+            try {
+                file_put_contents($path, $contents);
+                break;
+            } catch (\Exception $e) {
```
glensc: this block is nonsense, `file_put_contents` returns `FALSE` on error, not throws `Exception`

copitz: It isn't: file_put_contents not only returns false on error but triggers a warning - which will result in an error exception, which is just handled here:

    $> php -a
    file_put_contents("/test" ,"test");
    PHP Warning:  file_put_contents(/test): failed to open stream: Permission denied in php shell code on line 1

src/Composer/Satis/Builder/PackagesBuilder.php
copitz: Yes, I took it from [\Composer\Json\JsonFile::write()](https://github.com/composer/composer/blob/master/src/Composer/Json/JsonFile.php#L111) because that allows only to write to the path, the file was read from but we eventually need to write to another path (which was done by the ``rename`` construct before)

src/Composer/Satis/Builder/PackagesBuilder.php
glensc: there's no warnings to exception converter!
```
➔ php -r 'echo PHP_VERSION, "\n"; try { file_put_contents("/test" ,"test"); } catch (\Exception $e) { echo $e->getTraceAsString(); }'
5.6.21
PHP Warning:  file_put_contents(/test): failed to open stream: Permission denied in Command line code on line 1
PHP Stack trace:
PHP   1. {main}() Command line code:0
PHP   2. file_put_contents() Command line code:1
```

copitz: There should be - Symfony\Console should handle that. Please try to run the application with a denied target directory and you'll see

stof: @copitz Symfony Console does not register an error handler. It is not its job. However, Satis does it: https://github.com/composer/satis/blob/ca25ccddfc660894f083e1d1b5c11ed6271533f7/src/Composer/Satis/Console/Application.php#L51

PR #301, opened 2016-02-22
Regex fix for purging packages

ro0NL:
As discussed in https://github.com/composer/satis/pull/185#issuecomment-186236654


ro0NL:
:+1:

Issue #312, opened 2016-03-15
Adding vcs subversion support for target-branches, target-tags with multiple endpoints.


xylesoft:
We currently have a situation where a repository exists in the following structure:
```
src/
    branches/
        development/
        releases/
    tags/
        development/
        releases/
    trunk/
```

From my understanding the ```target-branches``` can only support a string to either ```branches/development```or ```branches/releases```, would it be feasible to add an array check on the target-branches configurations. For example:
```
{
    ...
    "target-branches": ["branches/releases","branches/development"],
    ...
}
```

Or is it possible to get all repository (containing a composer.json) end points recursively from just providing ```"taget-branches":"branches"``` ?

alcohol:
Shouldn't it be `branches-path`? Or are the docs horribly outdated?

https://getcomposer.org/doc/05-repositories.md#subversion-options

alcohol:
This seems to be the code that retrieves all branches from that path:

https://github.com/composer/composer/blob/15face5432d7b7334db6ac69fac0190971cafa6e/src/Composer/Repository/Vcs/SvnDriver.php#L230-L246

Perhaps you could specify the repository twice, and have a different `branches-path` for each instance? This does not seem like a suitable solution if you have many paths though. But it might suffice for just 2 or 3.

xylesoft:
I am currently using branches path, but I think perhaps I didn't clarify the structure well enough.

in the ```branches/development/``` directory, we also have other directories which are each of our development branches, so we have something like: ```branches/development/sto-new-feature```, which is what we work in. ```development``` and ```releases``` are just containers for a number of branches.

alcohol:
I understand that. I simply don't know what `svn ls` returns. I don't know how svn treats branches or determines them. So I'm not really in a position to provide much useful feedback here.

alcohol:
Hence my alternative solution;

```
repositories: [
  { "type": "vcs", "url": "http://svn.example.org/projectA/", "branches-path": "branches/development" },
  { "type": "vcs", "url": "http://svn.example.org/projectA/", "branches-path": "branches/releases" }
]
```

xylesoft:
Your solution appears to work fine, thank you for the heads up. Do you think submitting a pull request to allow an array of paths as initially suggested would be considered useful to the composer/satis project? If so I can look into it over this week at some point, certainly would save having to create do repository configuration objects.

glensc:
[...]

PR #320, opened 2016-04-01
`tags-path` and `branches-path` JSON schema update

xylesoft:
Implemented JSON schema update to validate the `tags-path` and `branches-path` configuration clauses, also enabled the possibility of using arrays for both path configurations.

This Pull Request relates to: https://github.com/composer/satis/issues/312 and https://github.com/composer/composer/pull/5142

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -95,7 +95,7 @@
                         "format": "uri"
                     },
                     "branches-path" : {
-                        "type": ["string","array"],
+                        "type": ["string","array","boolean"],
                         "description": "Single or array of relative paths to the Repository's Branches directories.",
                         "items": {
                             "type": "string",
@@ -107,7 +107,7 @@
                         "description": "Relative path to the Repository's Trunk directory."
                     },
                     "tags-path" : {
-                        "type": ["string","array"],
+                        "type": ["string","array","boolean"],
                         "description": "Single or array of relative paths to the Repository's Tags directories.",
                         "items": {
                             "type": "string",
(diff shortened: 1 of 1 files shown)


res/satis-schema.json
```diff
@@ -78,7 +78,44 @@
...
+            "additionalProperties": true,
+            "items": {
+                "type": "object",
+                "additionalProperties": false,
+                "required": ["type","url"],
+                "properties": {
+                    "type" : {
+                        "type": "string",
+                        "description": "Version control system type",
+                        "enum": ["vcs","svn","git","hg"]
```
stof: that's wrong. There are more type of repositories

stof: missing ones: ``composer`` and ``package`` at least.

stof:
``repositories`` should not have a strict validation in the Satis schema if it does not have it in the composer schema (as this config is about configuring composer itself)

stof:
The current PR will break the usage of Satis for many cases, as it misses lots of config

xylesoft:
Thank you for the feedback @stof, could you please give more some details of the missing areas regarding the config?

Regarding the `repositories`, which repositories are missing (besides composer and package)? I'm in the understanding that the schema should be an outline to the implementations permissible repositories.

alcohol:
https://getcomposer.org/doc/05-repositories.md#types

Not sure how completely up to date the documentation is, but it should contain everything.

res/satis-schema.json
alcohol: `artifact` also

alcohol:
Not sure what to do with this since your PR is also still open @ Composer.

/cc @Seldaek can you provide some feedback? See also referenced issue #5142

PR #323, opened 2016-04-05
Allow Symfony 3

nicoschoenmaker:
We're porting our applications to Symfony 3. Satis is used internally, and it currently depends on Symfony 2. The unit-tests, and the bin/satis command, seem to work as well with the `symfony/console` component of v3.

The only thing that did not work for me was the [platform](https://getcomposer.org/doc/06-config.md#platform) requirement to php 5.3.3, since Symfony v3 requires 5.5.9. Also, PHP 5.3 (or 5.4) are not getting security updates anymore, so it feels wrong to set it there.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "php": "^5.3.2 || ^7.0",
         "composer/composer": "1.0.*@dev",
         "twig/twig": "^1.7",
-        "symfony/console": "^2.1"
+        "symfony/console": "^2.1|^3.0.4"
     },
     "autoload": {
         "psr-0": { "Composer\\Satis": "src/" }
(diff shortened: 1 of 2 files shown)


alcohol:
Thanks :)

composer.json
```diff
@@ -34,10 +34,5 @@
         "phpunit/phpunit": "^4.5 || ^5.0.5",
         "phpunit/phpunit-mock-objects": "2.3.0 || ^3.0",
         "mikey179/vfsStream": "^1.6"
-    },
-    "config": {
-        "platform": {
-            "php": "5.3.3"
```
stof: Remove this entirely is wrong, as it means that the Satis phar won't be compatible with PHP 5.3 anymore (your lock file update bumped the requirement to PHP 5.5.9 as it uses Symfony 3)

glensc:
isn't config.platform root only setting? doesn't affect if it's in dependent package?

stof:
@glensc it is

PR #328, opened 2016-04-19
Add build command "stats" option to display a progress bar.

Divi:
This will hide all messages like "Dumping foo/bar..." to show a progress bar instead.

Example :

```
Scanning packages
Creating local downloads in 'web/dist'
  1/86 [>---------------------------]   1% - Installing foo/bar (dev-master)
```

The verbosity override the option : if the build command is executed with `-v(vv)` option, the `stats` option will be disabled.
Useful when having a lot of packages.

### DIFF
--- a/src/Composer/Satis/Builder/ArchiveBuilder.php
+++ b/src/Composer/Satis/Builder/ArchiveBuilder.php
@@ -14,6 +14,9 @@ namespace Composer\Satis\Builder;
 use Composer\Composer;
 use Composer\Factory;
 use Composer\Util\Filesystem;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * Builds the archives of the repository.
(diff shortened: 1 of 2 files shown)


src/Composer/Satis/Builder/ArchiveBuilder.php
```diff
@@ -53,15 +59,55 @@ public function dump(array $packages)
...
+        if ($isStats) {
+            $packageCount = 0;
+
+            foreach ($packages as $package) {
+                if (!$helper->isSkippable($package)) {
+                    ++$packageCount;
+                }
+            }
+
+            $this->composer->setDownloadManager($this->composer->getDownloadManager());
```
stof: what is the goal of this line ?

Divi: None, it's a mistake, maybe a wrong copy/paste. Didn't see it, thanks.

Should I create a new PR ?

alcohol: Yes :-).

Divi:
I'll do it tonight :)

src/Composer/Satis/Builder/ArchiveBuilder.php
Divi: I'll do it tonight :)

Issue #331, opened 2016-05-30
purge does not honor "skip-dev" setting


Adirelle:
`satis purge` removes dev packages although skip-dev is set to false in satis.json.
It causes all dev packages to be rebuilt on next `satis build`.

Satis version 1.0.0-dev

alcohol:
The `purge` command never claims to honor this setting. This setting does not belong to the purge command. I'll note this down as a feature request :-)

Adirelle:
I understand that, but right now if you enable building dev packages, they get deleted by purge, though they are referenced in the .json files.

alcohol:
Well this how `purge` was implemented. And since `satis` depends almost entirely on community contributions, all I can do is mark this as a feature request for you, and then hope that someone will pick it up :-)

alcohol:
I made some changes recently to the `purge` command. Can you see if this still behaves as expected? I think your old PR already resolved this issue, but would like to confirm it still works properly before closing this.

PR #332, opened 2016-05-31
Fixed the purge command.

Adirelle:
The purge command now extracts a strict list of files to keep from the JSON packages and removes all non-matching files from the archive directory. See #331.


alcohol:
Would it not be better to make this behaviour a flag on the `purge` command? I can imagine that someone might want the purge command to remove everything?

Adirelle:
> I can imagine that someone might want the purge command to remove everything?

There is `rm` for those cases.

alcohol:
Thanks.

PR #345, opened 2016-07-17
Changed package anchor, resolves #343

### LINKED ISSUE #343: Auto-hide package 'anchor' on the package title instead of appending an extra element
In response to #342.
PRs #172 and #302 added a little symbol to support page anchors. This is great however, in their current form, they add extra bloat to the page.

GitHub supports something similar on their wiki pages bij prefixing an anchor element with a fancy chain icon and hiding it by default. The anchor wil only pop up when the user is hovering over the title.

My suggestion would be to move the anchor to the front and do something similar as GitHub does with the hover states.

screen shot 2016-07-12 at 11 22 54

kevindierkx:


alcohol:
Thanks

PR #353, opened 2016-10-07
update dockerfile and templates

alcohol:
Would love to get some feedback on this from Satis users.


Dockerfile
```diff
@@ -1,16 +1,36 @@
...
+    && php -r " \
+        \$hash = hash('SHA384', file_get_contents('/tmp/composer-setup.php')); \
+        \$signature = trim(file_get_contents('/tmp/composer-setup.sig')); \
+        if (!hash_equals(\$signature, \$hash)) { \
+            unlink('/tmp/composer-setup.php'); \
+            echo 'Integrity check failed, installer is either corrupt or worse.' . PHP_EOL; \
+            exit(1); \
+        }"
+
+RUN php /tmp/composer-setup.php --quiet --ansi --install-dir=/usr/bin --filename=composer \
```
stof: this should be in the same layer than the download of the composer-setup.php file. Otherwise, it will still be part of the image layers

Dockerfile
```diff
@@ -1,16 +1,36 @@
...
+    && rm /tmp/composer-setup.php \
+    && composer --ansi --version
 
 WORKDIR /build
-ENTRYPOINT ["/satis/bin/satis"]
+WORKDIR /satis
```
stof: having 2 work dirs look wrong to me

README.md
```diff
@@ -11,18 +11,37 @@ to their version constraints) packages into a Composer Repository file.
 Usage
 -----
 
-- Download [Composer](https://getcomposer.org/download/): `curl -sS https://getcomposer.org/composer.phar -o composer.phar`
-- Install satis: `php composer.phar create-project composer/satis --stability=dev --keep-vcs`
+- Install satis: `php composer.phar create-project composer/satis:dev-master --keep-vcs`
```
stof: most projects are showing the usage of a global ``composer`` rather than ``php composer.phar``, especially when you remove the download step

views/index.html.twig
```diff
@@ -1,183 +1,118 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="robots" content="noindex,nofollow" />
-    <title>{{ name }} Composer Repository</title>
+
+    <title>{{ name }} Composer repository</title>
+
     <style type="text/css">
-    {% include 'styles.css' %}
+        {% include 'assets/bootstrap.min.css' %}
```
stof: changing the style should not be in the same PR than adding docker support IMO. they are 2 unrelated changes. Combining them in the same PR makes things much harder to review (and mean that none can be accepted until both are)

views/index.html.twig
```diff
@@ -1,183 +1,118 @@
...
+
+<div class="container">
+
+    <div class="page-header">
+        <h1>
+            {% if name %}{{ name }}{% else %}Composer repository{% endif %}
+        </h1>
+
+        <small id="lastupdate">
+            Last updated: <time datetime="{{ "now"|date(constant('\DateTime::ISO8601')) }}">{{ "now"|date(constant('\DateTime::RSS')) }}</time>
```
stof: should be updated to ``\DateTime::ATOM``, which is the real ISO format. ``\DateTime::ISO8601`` is not actually compatible with the ISO format

.dockerignore
```diff
@@ -0,0 +1 @@
+*
```
stof: what is this ignoring ?

Dockerfile
```diff
@@ -1,16 +1,36 @@
...
+
+RUN composer --ansi --quiet create-project composer/satis:dev-master .
```
stof: wouldn't it make more sense to copy the local code instead, and to run ``composer install`` ? This would ensure that it is built using the current codebase.
Due to docker caching, your existing setup will install the dev-master branch the first time you build the image, and never update the Satis codebase after that until you delete the docker cache (because this line will not change in the docker cache, and so cache will not be invalidated)

views/index.html.twig
alcohol: Fair enough. I think it should be reasonably easy to split this into 2 pull requests.

Dockerfile
alcohol: Correct. The image should contain the code as-is when the image is build. This creates a container which is essentially a deployable artifact.

If you want to run the container with your local development version of Satis, you can simply mount your local code on top of the path in which Satis is installed by using `-v $(pwd):/satis`.

README.md
alcohol: I can change that, no problem.

Dockerfile
alcohol: Alright, I can change that.

Dockerfile
alcohol: Alright, but then how can we make sure the user has run `composer install`? We need to verify that we are also `COPY`ing the `vendor/` directory into the image.

alcohol: I'll see what I can do, cheers.

glensc: the caching layers problem is usually solved with:

```
ADD composer.lock
RUN composer install
```

then it will use cache until `composer.lock` checksum is modified. same problem described here:
http://bitjudo.com/blog/2014/03/13/building-efficient-dockerfiles-node-dot-js/

README.md
```diff
@@ -52,16 +72,25 @@ Authors
 Jordi Boggiano - <j.boggiano@seld.be> - <http://twitter.com/seldaek> - <http://seld.be><br />
 Nils Adermann - <naderman@naderman.de> - <http://twitter.com/naderman> - <http://www.naderman.de><br />
 
-See also the list of [contributors](https://github.com/composer/satis/contributors) who participated in this project.
+See also the list of [contributors][] who participated in this project.
```
glensc: `[]` is superfluous, `[contributors]` works the same way

alcohol: Yeah it works on Github, but according to https://daringfireball.net/projects/markdown/syntax#link it should be `[label][]`.

PR #354, opened 2016-10-07
Docker template

alcohol:
@stof open to suggestions/modifications regarding the `COPY` vs `create-project`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+*
+!bin
+!res
+!src
+!views
+!composer.json
+!composer.lock
+!LICENSE
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,42 @@
-FROM php:5.6
+FROM php:7-alpine
 
-ENV DEBIAN_FRONTEND=noninteractive
+RUN apk --no-cache add curl git subversion openssh openssl mercurial tini
 
-RUN apt-get update \
-    && apt-get install -y git zlib1g-dev \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* \
-    && docker-php-ext-install zip \
-    && echo "date.timezone = UTC" > /usr/local/etc/php/php.ini
+RUN echo "memory_limit=-1" > $PHP_INI_DIR/conf.d/memory-limit.ini \
+ && echo "date.timezone=${PHP_TIMEZONE:-UTC}" > $PHP_INI_DIR/conf.d/date_timezone.ini
 
-RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin --filename=composer \
-    && composer create-project composer/satis --stability=dev --no-dev
+ENV COMPOSER_HOME /composer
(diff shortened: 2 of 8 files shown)


alcohol:
@stof does this look more like what you had in mind?

.dockerignore
```diff
@@ -0,0 +1,4 @@
+vendor/
+tests/
+.travis.yml
+phpunit.xml.dist
```
stof: you can also avoid including the readme, the box.json file and other config files. They are useless in the image. and ``/build`` should be ignored as well, in case you built a phar locally

Dockerfile
```diff
@@ -1,16 +1,43 @@
...
+ && composer --ansi --version
 
 WORKDIR /build
-ENTRYPOINT ["/satis/bin/satis"]
+WORKDIR /satis
+
+ADD ["composer.json", "composer.lock", "/satis/"]
+
+RUN composer --ansi install --no-dev \
+ && composer dump-autoload --optimize --no-dev
```
stof: dumping the autoload should be done after adding the Satis source, otherwise the satis source will not be autoloadable

alcohol: You are right, that makes more sense especially with the `--optimize` flag.

.dockerignore
alcohol: Will do.

alcohol:
@stof anything further?

Dockerfile
```diff
@@ -1,16 +1,42 @@
...
+ && rm /tmp/composer-setup.php \
+ && composer --no-ansi --version --no-interaction
 
-WORKDIR /build
-ENTRYPOINT ["/satis/bin/satis"]
+WORKDIR /satis
+
+ADD ["composer.json", "composer.lock", "/satis/"]
+
+RUN composer install --no-autoloader --no-ansi --no-dev --no-scripts --no-plugins --no-interaction --quiet
```
stof: do we need ``--quiet`` ? It prevents seeing errors

alcohol: We don't need it, no. I just could not find another way to skip all the "downloading" steps that were taking up most of the build output.

alcohol:
@Lance0312 are you still running Satis in a Docker container? If so, can you see if these changes work for you?

alcohol:
@stof if you have nothing further, I'll merge this afternoon and setup the image on Docker hub.

stof:
Looks good to me

Lance0312:
Yes, I'm still running Satis in a Docker container. It looks good to me with a quick review, and I'll give it a try ASAP.

composer.json
```diff
@@ -19,7 +19,7 @@
     ],
     "bin": ["bin/satis"],
     "require": {
-        "php": "^5.5 || ^7.0",
+        "php": "^5.6 || ^7.0",
```
alcohol: It's not a hard requirement. You can probably still get it to run if you do a `composer install|update --ignore-platform-reqs`. However, going forward we will no longer officially support EOL PHP versions for Satis.

PR #355, opened 2016-10-07
use bootstrap for theming/layout

alcohol:


GawainLynch:
Looks good here, tested on a Satis install with a couple of hundred repos … content seems to line up with what was expected, search works, an to these eyes at least looks much clearer.

PR #356, opened 2016-10-11
PHP Version Requirement In README

moebrowne:
The README currently states that the PHP requirement is 5.5+ when the actual requirement, as set in the `composer.json`, is 5.6+


stof:
:+1:

alcohol:
Thanks

PR #357, opened 2016-10-12
Docker tweaks

alcohol:
- COPY instead of ADD

  For other items (files, directories) that do not require ADD’s tar
  auto-extraction capability, you should always use COPY.

- add MAINTAINER

- double quote to prevent globbing and word splitting

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,14 +24,14 @@ RUN curl -s -f -L -o /tmp/composer-setup.php https://getcomposer.org/installer \
 
 WORKDIR /satis
 
-ADD ["composer.json", "composer.lock", "/satis/"]
+COPY ["composer.json", "composer.lock", "/satis/"]
 
 RUN composer install --no-interaction --no-ansi --no-autoloader --no-scripts --no-plugins --no-dev
 
-ADD /bin /satis/bin/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,14 +26,14 @@ RUN curl -s -f -L -o /tmp/composer-setup.php https://getcomposer.org/installer \
...
 RUN composer install --no-interaction --no-ansi --no-autoloader --no-scripts --no-plugins --no-dev
 
-ADD /bin /satis/bin/
-ADD /res /satis/res/
-ADD /views /satis/views/
-ADD /src /satis/src/
+COPY /bin /satis/bin/
+COPY /res /satis/res/
+COPY /views /satis/views/
+COPY /src /satis/src/
```
alcohol: Because the syntax does not allow for multiple destinations (only multiple sources).

glensc: but you are not renaming then, set destination to `/satis/` ?

stof: @glensc if the source is a directory, Docker copies the *content* of the directory, not the directory itself. So your suggestion would not work

alcohol: Yeah, what @stof said.

PR #358, opened 2016-10-17
template + javascript changes

alcohol:
* drop table structure, use div.row + div.col-*
* remove javascript window.location.hash hooks
* clear search/filter when pressing "ESC"
* drop last updated from footer

### DIFF
--- a/views/index.html.twig
+++ b/views/index.html.twig
@@ -34,8 +34,8 @@
             {% if name %}{{ name }}{% else %}Composer repository{% endif %}
         </h1>
 
-        <small id="lastupdate">
-            Last updated: <time datetime="{{ "now"|date(constant('\DateTime::ISO8601')) }}">{{ "now"|date(constant('\DateTime::RSS')) }}</time>
+        <small>
+            Last updated: <time datetime="{{ "now"|date(constant('\DateTime::RFC3339')) }}">{{ "now"|date('l, d M Y H:i:s T') }}</time>
         </small>
 
         {% if description %}<p>{{ description|raw }}</p>{% endif %}
@@ -53,13 +53,11 @@
     </div>
 
     <div id="package-list">
-        {% for name, package in packages %}
-            {% include 'package.html.twig' %}
-        {% endfor %}
+        {% for name, package in packages %}{% include 'package.html.twig' %}{% endfor %}
     </div>
 
     <div id="ft">
-        <p><small>This Composer repository is powered by <a href="https://github.com/composer/satis">Satis</a> {{ constant('Composer\\Satis\\Satis::VERSION') }} - Last updated:  {{ "now"|date(constant('\DateTime::RFC3339')) }}</small></p>
+        <p class="text-center"><small>This Composer repository is powered by <a href="https://github.com/composer/satis">Satis</a> {{ constant('Composer\\Satis\\Satis::VERSION') }}</small></p>
     </div>
 
 </div>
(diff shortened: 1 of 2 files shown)


views/index.html.twig
```diff
@@ -34,8 +34,8 @@
             {% if name %}{{ name }}{% else %}Composer repository{% endif %}
         </h1>
 
-        <small id="lastupdate">
-            Last updated: <time datetime="{{ "now"|date(constant('\DateTime::ISO8601')) }}">{{ "now"|date(constant('\DateTime::RSS')) }}</time>
+        <small>
+            Last updated: <time datetime="{{ "now"|date(constant('\DateTime::RFC3339')) }}">{{ "now"|date(constant('\DateTime::RFC3339')) }}</time>
```
stof: the content of the tag should use a more readable representation than the RFC one (it is meant to contain a human-readable representation)

views/package.html.twig
```diff
@@ -30,79 +30,76 @@
         <p class="lead">{{ package.highest.description }}</p>
         {% endif %}
 
-        <table class="table table-striped">
+        {% if package.highest.homepage %}
+        <div class="row">
+            <div class="col-lg-2 text-right"><strong>Homepage</strong></div>
+            <div class="col-lg-10"><a href="{{ package.highest.homepage }}">{{ package.highest.homepage }}</a></div>
```
stof: Using ``col-lg-*`` without having classes for the lower resolutions creates a weird thing, because ``.row`` has a negative horizontal margin to compensate the grid gutter of columns, but your divs don't create these gutters.

If you always want this on 2 columns, use ``col-xs-*`` (grids are bottom-up in bootstrap. ``col-xs`` will apply to higher resolutions as long as you don't overwrite it). Otherwise use ``col-xs-12``. And in any case, it can use 2 columns before reaching ``-lg-``

src/Composer/Satis/Builder/PackagesBuilder.php
```diff
@@ -29,6 +29,7 @@ class PackagesBuilder extends Builder
     /** @var string included json filename template */
     private $includeFileName;
 
+    /** @var array */
```
alcohol: Damnit, I thought I skipped over this when I hit `git commit -p`. Hold on.

views/index.html.twig
alcohol: Ok, I would argue that `DATE_COOKIE` is more readable than `DATE_ISO8601` though. What is your thought on this?

Also, as far as I understand it, this puts the "format" in the tag, not the value (e.g. `l, d-M-Y H:i:s T` in the case of `COOKIE`).

stof: you are using RFC3339 for both the machine-readable representation (fine) and the human-readable one (less fine) here. There is no Date_COOKIE involved here

alcohol: Ah wait, no, that's the value passed to the twig date function. My bad.

Still though, thoughts on format?

alcohol: Actually according to this, RFC3339 is the valid format: https://www.w3.org/TR/html-markup/datatypes.html#common.data.datetime

views/package.html.twig
alcohol: Yeah I just realized that the responsive alignment classes only exist in bootstrap 4 :<

Maybe use a dl + dd/dt approach instead?

stof: please add the proper markdown code blocks to make your comment readable.

views/index.html.twig
stof: Well, the existing DATE_RSS is fine too, but the full day name of DATE_COOKIE is indeed more readable. On the other hand, DATE_COOKIE displays ``12-Aug-2013`` while DATE_RSS uses ``12 Aug 2013``, so it would win there

alcohol: Used a combination of both.

PR #359, opened 2016-10-19
compiler is out-of-date compared to box.json

alcohol:
box is quite elegant and it makes little sense to have to keep two
different approaches constantly in sync


stof:
:+1:

PR #360, opened 2016-10-19
simplify dir structure and namespaces

alcohol:


composer.json
```diff
@@ -25,10 +25,10 @@
         "symfony/console": "^2.1 || ^3.0.4"
     },
     "autoload": {
-        "psr-0": { "Composer\\Satis": "src/" }
+        "psr-4": { "Composer\\Satis\\": "src" }
     },
     "autoload-dev": {
-        "psr-0": { "Composer\\Test\\Satis": "tests/" }
+        "psr-4": { "Composer\\Satis\\": "tests" }
```
stof: IMO, we should keep tests in a different namespace than the code itself

alcohol: I never really understood the purpose of that, especially since we have them in `require-dev`.

PR #361, opened 2016-10-19
Fix template for x-small screens

aaukt:
Like @stof suggested, use col-xs-12 so the column will wrap onto a new line. For small screens switch back to two columns next to each other.


alcohol:
I am guessing we should not use `text-right` on x-small screens though. Maybe switch to bootstrap 4, since 3 does not yet support responsive alignment classes :(

alcohol:
I made some additions. Feedback please?

PR #362, opened 2016-10-19
cleanup, update, etc

alcohol:
Afaik, webmozart/assert is only needed by annotations lib from phpdocumentor, which we dont use in the phar.


stof:
Rather than excluding phpunit and other stuff from box.json, which is error-prone to maintain, we should build the phar from a ``--no-dev`` install. It would avoid bring unwanted stuff in the phar if PHPunit adds a new dependency for instance.

alcohol:
Well.. we don not really build the Phar explicitly anywhere. It is only there for people who want to do that. Maybe we should build the Phar for the Docker container? Not sure if it is better than just dumping all the source code into the container. Open to suggestions or more feedback.

alcohol:
Also, I do not really know how we could enforce the `--no-dev` environment from a box.json configuration perspective.

stof:
@alcohol note that having the dev requirements in the phar is not breaking it. It just makes it bigger than necessary.

stof:
and IMO, we should build the phar for releases and attach it to github releases

alcohol:
Alright, so then we are manually in control of this.

PR #363, opened 2016-10-20
split dist files per vendor and package

### LINKED ISSUE #68: dist files could follow the composer cache organization
For the sake of clarity, and to match the composer cache, dist files could be relocated to something like `output-dir/directory/vendor-name/bundle-name/`

The vendor and bundle parts could also be dropped in the file name .

alcohol:
closes #68

implementation is kind of cringe-worthy, but so is most of the codebase unfortunately, so not sure if worth investing more effort

### DIFF
--- a/src/Builder/ArchiveBuilder.php
+++ b/src/Builder/ArchiveBuilder.php
@@ -97,13 +97,16 @@ class ArchiveBuilder extends Builder
                     $this->output->setVerbosity(OutputInterface::VERBOSITY_QUIET);
                 }
 
+                $intermediatePath = preg_replace('#[^a-z0-9-_/]#i', '-', $package->getName());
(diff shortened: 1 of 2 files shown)


stof:
:+1:

alcohol:
I tested it quickly with a very basic satis configuration. Worked as expected, but no idea if there might be edge-cases.

aaukt:
Correct me if i'm wrong, but i think it should be noted that people should be cautious with rolling this out to existing satis repositories, because the old paths are still in the composer.lock.

So either do not delete the old archives or put some clever redirect logic in place.

alcohol:
Nothing gets deleted by default. Deletions only happen when you run the purge command.

aaukt:
I know, just if you wonder why you should keep a duplicate of all archives.

PR #364, opened 2016-10-22
Fix fatal errors

jyggen:
Fixes the following fatal errors:
```shell
PHP Fatal error:  Cannot use Symfony\Component\Console\Input\InputArgument as InputArgument because the name is already in use in /path/to/satis/src/Console/Command/PurgeCommand.php
```
```shell
PHP Fatal error:  Cannot use Composer\Json\JsonFile as JsonFile because the name is already in use in /path/to/satis/src/Console/Command/PurgeCommand.php
```


alcohol:
Shame on me :(

Not sure why travis did not pick up on this though.

alcohol:
Regardless, thanks.

PR #365, opened 2016-10-23
Remove duplicate line

anoopmd:
Fixes : 
```php
PHP Fatal error:  Cannot use Symfony\Component\Console\Input\InputArgument as InputArgument because the name is already in use in src/Console/Command/PurgeCommand.php on line 19
```


alcohol:
Thanks, was fixed in #364 as well though :-)

Issue #367, opened 2016-11-04
"providers": true will break installation of packages with a replace package definition


andygrunwald:
We run our own instance of [Medusa](https://github.com/instaclick/medusa) with the configuration:

```
{
    "repositories": [],
    "require": [
        "symfony/symfony",
        "symfony/filesystem"
    ],
    "repodir": "/appdata/git-mirror",
    "satisurl": "http://my.url/git-mirror",
    "satisconfig": "satis.json"
}
```

So Medusa is mirroring the repositories.
After this is done we kickoff satis do generate the usual Satis html with the configuration:

```
{
    "name": "php package mirror",
    "homepage": "http://my.url/external/",
    "require-all": true,
    "require-dependencies": true,
    "providers": true,
    "archive": {
        "directory": "dist",
        "format": "tar",
        "prefix-url": "http://my.url/external/",
        "skip-dev": true
    },
    "repositories": [
        ...
    ]
}
```

The important part here is `"providers": true,`.
The providers feature was introduced in #295.

If `providers` is set to `true` this composer json will not be able to install both packages and leads to a conflict.

```
{
    "name": "foo/test",
    "authors": [
        {
            "name": "foo",
            "email": "foo@bar.com"
        }
    ],
    "repositories": [
        {
            "packagist": false
        },
        {
            "type": "composer",
            "url": "http://my.url/external"
        }
    ],
    "require": {
           "symfony/filesystem": "3.1.6",
           "symfony/symfony": "3.1.6"
    }
}
```

This composer.json will succeed, but it will install filesystem in 3.1.5 and symfony in 3.1.6 or vice versa:
```
{
    "name": "foo/test",
    "authors": [
        {
            "name": "foo",
            "email": "foo@bar.com"
        }
    ],
    "repositories": [
        {
            "packagist": false
        },
        {
            "type": "composer",
            "url": "http://my.url/external"
        }
    ],
    "require": {
           "symfony/filesystem": "^3.0",
           "symfony/symfony": "^3.0"
    }
}
```

The reason is the [usage of replace in symfony](https://github.com/symfony/symfony/blob/master/composer.json#L30).
@aaukt [mentioned this in the pull request](https://github.com/composer/satis/pull/295#issuecomment-220902213).

> From having build this myself i can tell you, that special care has to be taken of "replace" packages. The package that is replaced has to include the version information of its "replacee" in the json file.
[...]

PR #371, opened 2016-11-18
Include replacing packages into package providers (#367)

copitz:
Please see #367 for a detailed explanation on what this change fixes

### DIFF
--- a/src/Builder/PackagesBuilder.php
+++ b/src/Builder/PackagesBuilder.php
@@ -70,12 +70,18 @@ class PackagesBuilder extends Builder
             }
             $repo['providers'] = [];
             $i = 1;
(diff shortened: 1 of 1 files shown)


alcohol:
Thanks

stof:
this is missing support for ``provide`` though

copitz:
You're right - the thing is that I didn't know how to reflect provided packages in the provider includes. If s.o. could help me figuring this out, we/I could handle this in a new issue...

stof:
@copitz hmm, Packagist indeed does not dump a provider file for packages which are only provided. But this looks like a bug in Packagist IMO, as it means that composer will not be able to find packages providing ``psr/http-message-implementation`` for instance

Issue #374, opened 2016-12-06
[Question/Feature Request] Display required packages in web view


mooror:
Hey there everyone,
I was wondering if the web builder can be setup to output a "Requires" row for each item. I see that you have a "Required By" row already, but I find it hard to pinpoint an items requirements using is system (a lot of searching through the page). Normally I would just go into the composer.json file for that item and look up its requirements, but my satis system appears to be missing some of those requirements (an issue I'm also trying to fix). I don't mind adding this functionality myself if need be, but I want to be as time efficient as I can and therefor I thought I would come to you first.

Thanks in advance,
Mooror

mooror:
I went ahead and implemented this feature for my install. I also created a [pull request](https://github.com/composer/satis/pull/375) to hopefully help merge these changes into the core.

PR #375, opened 2016-12-06
[Issue #374] Added row for displaying package requirements 

mooror:
I went ahead and implemented the requirements row. Now I found that my version of satis is an older one so I actually had to adapt this code for my installation. I have not tested this code with the new version and therefor cannot confirm that it works. My hope is that it either works without issue or will give you a bit of a head start.

Hope it helps a bit,
Mooror

P.S. For the record, I really enjoy what satis does but I find the name rather unsettling.

### DIFF
--- a/src/Builder/WebBuilder.php
+++ b/src/Builder/WebBuilder.php
@@ -27,6 +27,9 @@ class WebBuilder extends Builder
 
     /** @var PackageInterface[] List of calculated required packages. */
     private $dependencies;
+    
+    /** @var PackageInterface[] List of packages that an package relies on */
+    private $requirements;
 
     /** @var \Twig_Environment */
     private $twig;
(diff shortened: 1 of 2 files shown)


alcohol:
I tried it out, but I am not sure this adds much value. It attempts to link to anchors on the same page that do not necessarily exist. Nor is it possible to determine which constraints apply. Also, the requirements might not actually apply to all listed versions.

satis

mooror:
I was aware of the problem with the links but I had overlooked the second problem you mentioned...For the first issue I guess we could just forgo the <a> tags all together and just list the items as direct children of the <li> tags.

mooror:
I am curious as to how my requirements row differs from the concept behind the "Required By" row. I might have missed it in the code, but I don't recall seeing version handling for that row either. If this is indeed the case then I would argue that just as the "Required By" row lets the developer see packages that rely on the item, regardless of the version; my row lets the developer know that one of the versions of the item relies on packages in the list.

Looking forward to more discussion,
Mooror

P.S. Sorry if my post is a bit confusing

mooror:
As I discovered in issue #376, my requirements row implementation seems to be unreliable. Unfortunately I don't have time to look into this more but I would definitely recommend that someone do a full test on my code before thinking about merging it.

Sorry for the let down,
Mooror

PR #378, opened 2016-12-06
Fix the disabling of Packagist

stof:
The key in the array was renamed in Composer at some point (and this array is not part of the public API of composer covered by semver, as any programmatic API currently)

### DIFF
--- a/src/Console/Command/BuildCommand.php
+++ b/src/Console/Command/BuildCommand.php
@@ -158,7 +158,7 @@ EOT
         }
 
         // disable packagist by default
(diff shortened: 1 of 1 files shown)


alcohol:
Yeah was thinking about this too. But then again, if you just update the .lock file with the latest versions, then you wont need the old key anymore. So.. not sure.

stof:
@alcohol but I know several people installing Satis as a dependency (for instance with ``composer global require composer/satis``) and so not relying on the lock file.

alcohol:
Yeah, either way, this change is fine :-)

PR #387, opened 2017-01-03
Added possibility to include the VCS directories when creating dist package

meridius:
Works by adding the `ignore-vcs` bool option to the `archive` section. This option defaults to **true**, so VCS directories are ignored by default, hence original behavior is preserved.

This is extremely useful when working with bleeding edge of packages and/or in need to hot-switch the committed states of required package, whilst **preserving** the ability to cache the dist package and rapidly **speed-up** the install time.

Depends on composer/composer#6035

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,9 +21,15 @@
         "irc": "irc://freenode.net/composer"
     },
     "bin": ["bin/satis"],
+    "repositories": [
+        {
+            "type": "vcs",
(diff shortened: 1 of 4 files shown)


glensc:
https://github.com/composer/composer/pull/6035 was rejected, so perhaps close this one here too?

alcohol:
Since it depends on that functionality, that makes sense yes.

PR #392, opened 2017-01-19
prevent load old version if it's not required

pcabreus:
I had a problem when I wanted to have a local copy of other repositories e.g. `"symfony/symonfy: ">=2.7"`. But I also wanted all dependencies like `"doctrine/common"` so I configured `"require-dependencies": true` and everything work fine. 

What was the problem with this? Well satis downloaded all version of everything, it's like I was set "require-all": true. I don't wanted old versions, I only wanted the last version after 2.7, not 2.0, 2.1 ....

So with this pull request I try to improve this feature. You can now download right versions. Also if you set for example `"doctrine/common": ">=2.5"`, it will download `>=2.5` as we expected and not `~2.4` that require symfony, because `>=2.5` match well with `~2.4`.

This work fine for me, but I only have this case of use, please check it.

### DIFF
--- a/src/Console/Command/BuildCommand.php
+++ b/src/Console/Command/BuildCommand.php
@@ -47,6 +47,7 @@ class BuildCommand extends BaseCommand
                 new InputOption('repository-url', null, InputOption::VALUE_OPTIONAL, 'Only update the repository at given url', null),
                 new InputOption('no-html-output', null, InputOption::VALUE_NONE, 'Turn off HTML view'),
                 new InputOption('skip-errors', null, InputOption::VALUE_NONE, 'Skip Download or Archive errors'),
+                new InputOption('best-candidate-strategy', null, InputOption::VALUE_NONE, 'Download only the best candidate version'),
                 new InputOption('stats', null, InputOption::VALUE_NONE, 'Display the download progress bar'),
             ])
             ->setHelp(<<<'EOT'
@@ -113,6 +114,7 @@ EOT
         $packagesFilter = $input->getArgument('packages');
         $repositoryUrl = $input->getOption('repository-url');
         $skipErrors = (bool) $input->getOption('skip-errors');
+        $bestCandidateStrategy = (bool) $input->getOption('best-candidate-strategy');
 
         // load auth.json authentication information and pass it to the io interface
         $io = $this->getIO();
@@ -172,7 +174,7 @@ EOT
         /** @var $application Application */
         $application = $this->getApplication();
         $composer = $application->getComposer(true, $config);
-        $packageSelection = new PackageSelection($output, $outputDir, $config, $skipErrors);
+        $packageSelection = new PackageSelection($output, $outputDir, $config, $skipErrors, $bestCandidateStrategy);
 
         if ($repositoryUrl !== null) {
             $packageSelection->setRepositoryFilter($repositoryUrl);
--- a/src/Console/Command/BuildCommand.php
+++ b/src/Console/Command/BuildCommand.php
@@ -47,7 +47,6 @@ class BuildCommand extends BaseCommand
                 new InputOption('repository-url', null, InputOption::VALUE_OPTIONAL, 'Only update the repository at given url', null),
                 new InputOption('no-html-output', null, InputOption::VALUE_NONE, 'Turn off HTML view'),
                 new InputOption('skip-errors', null, InputOption::VALUE_NONE, 'Skip Download or Archive errors'),
-                new InputOption('best-candidate-strategy', null, InputOption::VALUE_NONE, 'Download only the best candidate version'),
                 new InputOption('stats', null, InputOption::VALUE_NONE, 'Display the download progress bar'),
             ])
             ->setHelp(<<<'EOT'
@@ -114,7 +113,6 @@ EOT
         $packagesFilter = $input->getArgument('packages');
         $repositoryUrl = $input->getOption('repository-url');
         $skipErrors = (bool) $input->getOption('skip-errors');
-        $bestCandidateStrategy = (bool) $input->getOption('best-candidate-strategy');
 
         // load auth.json authentication information and pass it to the io interface
         $io = $this->getIO();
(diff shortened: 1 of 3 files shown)


alcohol:
Do you have a simple satis json with 2 or 3 libraries that can easily demonstrate the difference between before and after patch?

pcabreus:
In my last commit I changed the strategy, because what I want is to download only the best candidate of the library I need. So I passed from GBs to a few MBs of my static composer.

src/Console/Command/BuildCommand.php
```diff
@@ -47,6 +47,7 @@ protected function configure()
                 new InputOption('repository-url', null, InputOption::VALUE_OPTIONAL, 'Only update the repository at given url', null),
                 new InputOption('no-html-output', null, InputOption::VALUE_NONE, 'Turn off HTML view'),
                 new InputOption('skip-errors', null, InputOption::VALUE_NONE, 'Skip Download or Archive errors'),
+                new InputOption('best-candidate-strategy', null, InputOption::VALUE_NONE, 'Download only the best candidate version'),
```
stof: IMO, this should be part of the config file, not of the CLI flags.

src/PackageSelection/PackageSelection.php
```diff
@@ -186,6 +193,27 @@ public function select(Composer $composer, $verbose)
...
+                            $newConstraint,
+                            $link->getDescription(),
+                            $newConstraint->getPrettyString()
+                        );
+                    }
+                    unset($links[$key]);
+                }
+            }
+            /** @var Link[] $links */
+            $links = array_merge($links);
```
stof: merging a single array ? Why ? If you want to re-index it, use ``array_values`` instead. It would be more more understandable

src/PackageSelection/PackageSelection.php
```diff
@@ -461,4 +491,26 @@ private function filterRepositories(array $repositories)
...
+        try {
+            if (!$this->bestCandidateStrategy) {
+                return $pool->whatProvides($name, $link->getConstraint(), true);
+            } else {
+                $selector = new VersionSelector($pool);
+                if ($best = $selector->findBestCandidate($name, str_replace(['[', ']'], ['', ''], $link->getConstraint()->getPrettyString()))) {
+                    return [$best];
+                }
+            }
+        } catch (\Exception $e) {
```
stof: why this try/catch ?

src/PackageSelection/PackageSelection.php
```diff
@@ -461,4 +491,26 @@ private function filterRepositories(array $repositories)
...
+     * @return array|\Composer\Package\PackageInterface[]
+     */
+    private function findMatches(Pool $pool, $name, Link $link)
+    {
+        try {
+            if (!$this->bestCandidateStrategy) {
+                return $pool->whatProvides($name, $link->getConstraint(), true);
+            } else {
+                $selector = new VersionSelector($pool);
+                if ($best = $selector->findBestCandidate($name, str_replace(['[', ']'], ['', ''], $link->getConstraint()->getPrettyString()))) {
```
stof: passing the pretty string to the selector, with replacements done in it is a bad idea IMO (and very fragile)

src/PackageSelection/PackageSelection.php
pcabreus: I'm having some Exception with some libraries that has a wrong version like ">=3.0<4" or "7.0.0-7.0.2" or something like that. Composer doesn't recover from that problem so I have to do it for myseft. I think it's a Composer not Satis problem.

src/PackageSelection/PackageSelection.php
pcabreus: That is why I propose this solution here, I couldn't find other way to have the right version, so I had to clean it. And right now it's working

src/PackageSelection/PackageSelection.php
pcabreus: You're right

src/Console/Command/BuildCommand.php
pcabreus: I agree, but I was looking for a simple way to do it. The config file scare me a little and I don't know if it's a good idea or not add more configuration to the config file.

stof: it is better than adding more CLI flags that you must pass each time you build to get the same output (btw, this flag would also impact the prune command, which is also why having it in the config file is better)

src/PackageSelection/PackageSelection.php
stof: well, this is a separate topic than restricting versions IMO (unless this happens only in the VersionSelector, in which case the try/catch should wrap only this part of the code)

pcabreus:
For example if you require: `"symfony/form": "~2.7.11||~2.8.4||~3.0.4"`, and you put in your config file `"require-dependencies": true,` you will download more than 637 packages depending on how many new version has each dependencies. But if you put also the new config `"require-best-dependencies": true` you will download only 20 packages and it should be like this for a long time untill some packages add a new dependency. Every time you update your repos you will update its versions instead of adding more vertions.

src/Console/Command/BuildCommand.php
pcabreus: Done!!!

src/PackageSelection/PackageSelection.php
pcabreus: I will follow https://github.com/composer/semver/issues/54 to see what happen

alcohol:
I think "best" is a bit ambiguous. Give it a few months and I won't remember anymore what it actually means, which is more in the line of "limit each package to the latest version that matches their given constraint".

pcabreus:
"best" remember me "findBestCandidate", which is the function I use to do it. Any recomendation is welcome.

simesy:
I think this can be closed with https://github.com/composer/satis/pull/549

pcabreus:
I agree, I've been using it so far, just as a composer copy to work offline. But I just moved to Barcelona with a very good connection so I won't need it. 
Closed.

simesy:
Thanks for your code, i learned from it to do my PR

PR #400, opened 2017-02-26
Small layout improvements, optimized SVG usage and 'fixed' panel-titles

kevindierkx:
@glensc Mentioned in #343 that the SVG output would be repeated for every package, this is now replaced with SVG symbol/use attributes reducing the actual output.

Also the panel titles are rather large since the introduction of bootstrap, I've added the .panel-title class to resolve this issue and fix some alignment stuff.


alcohol:
Thanks

PR #411, opened 2017-04-25
Add zip extension to Docker image

titpetric:
The zip extension is notably missing from the Docker image. If one want's to use an artifact repository, the following exception is produced.

~~~
  [RuntimeException]
  The artifact repository requires PHP's zip extension
~~~


alcohol:
Merged in a891219e0ed70d82a378e61c699e0e7be0a5abba

titpetric:
Thanks for merging this in so fast! Works like a charm ;)

PR #414, opened 2017-05-04
add missing libraries required to clone repositories with SSH uri

pendexgabo:
Given the changes introduced in the commit https://github.com/composer/satis/commit/f4ce516bb82e7d73fa70695b68f7a5306b3bdd2a the `ssh` command will be missing thus, git actions involving a SSH uris will result on errors: `error: cannot run ssh: No such file or directory `

This pull request only adds those two missing libraries.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM php:7-alpine
 
 MAINTAINER https://github.com/composer/satis
 
-RUN apk --no-cache add curl git subversion mercurial tini zlib-dev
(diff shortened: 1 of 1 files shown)


alcohol:
Does SSH really need openssl installed? Would it not automatically install it as a requirement in that case? Please, provide some evidence supporting that both packages are required.

alcohol:
Added back in 146ad270b0d7976d0d5dfeb73701a2fa62a1c791.

pendexgabo:
thank you.

PR #429, opened 2017-08-01
[Feature] Save search string in url hash and load on pageload

Jack992:
Within this change we write every user input to the hash.
Also we read the hash on initial page-load and execute a search with it.
This makes it possible to send links with the search term.
Also when reloading the page to check if satis already rebuild, the search stays.


Jack992:
Failing checks are not caused by my changes :-)

alcohol:
Thanks

PR #431, opened 2017-08-09
Adding Support for GitHub as a Satis Repository

mikeschinkel:
This simple change allows generating a `packages.json` that will work when hosted in a Git repository.  

This is just a few lines of code that adds a `providers-prefix` property within `.archive` in `satis.json`. I have a working example of this currently located at [github.com/newclarity/wordpress-dependencies](https://github.com/newclarity/wordpress-dependencies/) which only requires this additional setting in `satis.json` _(I am using pseudo-code here):_

```.archive.providers-prefix="master"```

This allows generating a `/p/` URL with `/master/p/` which is what is needed for hosting the Composer repo as a Git repo.

The reason for this is my client did not want to host additional infrastructure and did not want to depend on anything other that GitHub for being able to rebuild and deploy their solution.

**NOTE:** My repo currently depends on [a CircleCI script to change `/p/` to `/master/p/`](https://github.com/newclarity/wordpress-dependencies/blob/master/scripts/compile.sh#L47) so my `satis.json` does not have  `providers-prefix`, but I would really like to see it be possible to run Satis and have it work without have to run scripts to modify its output.

**Also No guarantee that my example Git repo will be there long term _(because it is for a client)_, so please review this sooner than later?**

Thanks in advance for considering.

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -69,6 +69,10 @@
                     "description": "Location of the downloads.",
                     "format": "uri"
                 },
+                "providers-prefix": {
+                    "type": "string",
+                    "description": "Prefix for the provider URI template"
+                },
                 "checksum": {
                     "type": "boolean",
                     "description": "If false, will not provide the sha1 checksum for the dist files."
--- a/src/Builder/PackagesBuilder.php
+++ b/src/Builder/PackagesBuilder.php
@@ -26,7 +26,10 @@ class PackagesBuilder extends Builder
     /** @var string packages.json file name. */
     private $filename;
 
-    /** @var string included json filename template */
+	/** @var string prefix for providers URL. */
+	private $providersPrefix = null;
+
+	/** @var string included json filename template */
     private $includeFileName;
 
     private $writtenIncludeJsons = [];
@@ -44,7 +47,8 @@ class PackagesBuilder extends Builder
         parent::__construct($output, $outputDir, $config, $skipErrors);
 
         $this->filename = $this->outputDir . '/packages.json';
-        $this->includeFileName = isset($config['include-filename']) ? $config['include-filename'] : 'include/all$%hash%.json';
+	    $this->includeFileName = isset($config['include-filename']) ? $config['include-filename'] : 'include/all$%hash%.json';
+	    $this->providersPrefix = isset($config['archive']['providers-prefix']) ? trim( $config['archive']['providers-prefix'], '/' ) . '/' : '';
     }
 
     /**
@@ -62,7 +66,7 @@ class PackagesBuilder extends Builder
 
         $repo = ['packages' => []];
         if (isset($this->config['providers']) && $this->config['providers']) {
-            $providersUrl = 'p/%package%$%hash%.json';
+            $providersUrl = $this->providersPrefix . 'p/%package%$%hash%.json';
             if (!empty($this->config['homepage'])) {
                 $repo['providers-url'] = parse_url(rtrim($this->config['homepage'], '/'), PHP_URL_PATH) . '/' . $providersUrl;
             } else {


stof:
If you want to use a github repo to host this, shouldn't you be using the github pages thing to host the static site ?

curry684:
Indeed, seems like feature creep for something that already has a native alternative.

mikeschinkel:
@stof:
> _"If you want to use a github repo to host this, shouldn't you be using the github pages thing to host the static site ?"_

I did not try GitHub pages because 1.) I disabled the `index.html` generation because I did not need nor want for my use-case so making the idea of a GitHub Page website counterintuitive, 2.) I have never used GitHub Pages so was not familiar with how they work, 3.) I found the behavior of the `satis.json` properties `.output-dir`, `.archive.directory`, `.archive.absolute_directory` counterintuitive and confusing so I was unsure I would be able to get the URLs to work with GitHub pages, and 4.) having to set up GitHub pages seemed like additional setup for something I wanted people to be abled to just fork and have it work for them _(or do GitHub pages come along with a fork?)_.

That said, if GitHub pages *_are_* a viable option I can certainly explore that as an alternate to the approach proposed in my pull request.   Do you know for a fact that GitHub pages can be used in this way, or were you assuming they could?  If the former, can you point me to a repo that is already doing it?

Thanks in advance.

curry684:
GitHub Pages boils down to making a `gh-pages` branch in your project which is then automatically published as a static website. Since Satis does nothing more than publish static data that's essentially all you need.

- [Starter guide on how to publish on Pages](https://help.github.com/articles/user-organization-and-project-pages/)
- [Using HTTPS on GitHub pages](https://help.github.com/articles/securing-your-github-pages-site-with-https/)
- [Custom domains with GitHub pages](https://help.github.com/articles/using-a-custom-domain-with-github-pages/)

MatthiasKuehneEllerhold:
You're limited to one github pages instance per account in the free model, right?

curry684:
No, one user instance. You can have unlimited project instances.

Also, nobody's stopping you from publishing multiple Satis repositories in a single Pages instance.

stof:
and you can have github pages for any repo

mikeschinkel:
@curry684 

Thanks.  I will check it out and report back here my results, pro or con _(as soon as I have a chance; my repo/CircleCI build works as is and my client has asked me to move on the next step in the project.)_

mikeschinkel:
@curry684 @stof Hi guys.  Thanks for your suggestion.  

For one of my use-cases GitHub pages works _excellently_.  I created a public project on GitHub called [_Composer Dependencies for WordPress_](https://github.com/wplib/wp-composer-dependencies) that services two purposes:

1. It is an example and tutorial for how to quickly and easily create a GitHub-hosted Composer repo using Satis and CircleCI.
2. It is also a working Composer repo containing some of the plugins and themes that my team and/or some of client's websites use.  

You can see this projects GitHub Pages [**here**](https://wplib.github.io/wp-composer-dependencies).
 
**Unfortunately however** GitHub Pages **will not work** for the more likely use-case of a `private` Composer repo on GitHub. As you can see a [**GitHub Pages website cannot be made `private`**](https://stackoverflow.com/a/28745353/102699).  And that is the use-case my client essentially hired me to build.  

So the following is actually not true:

> _... for something that already has a native alternative._

I can of course workaround this by adding a Bash+JQ step to modify the generated `packages.json` file. But that feels like a very unfortunate requirement. Do you really Satis to be unable to natively generate the URL structures needed for very convenient, low effort and reliable repository hosting? Do you really want to leave post-processing of Satis-generated files as the only option?

Revote please?

curry684:
> the following is actually not true

Well it was until you changed the requirements to have the repository be private   😉 

Think @alcohol should have a looksie here.

mikeschinkel:
> _"Well it was until you changed the requirements to have the repository be private 😉"_

Fair point.  :-)   

Of course I didn't realize I needed to specify, I just assumed. Then I didn't realize GitHub Pages couldn't be private until I went looking for how to make them private! :-o

The good news is GitHub Pages are a great solution for public ones, so definite thanks for that.

alcohol:
Considering how small the change is, I am fine with merging this. But it will be Monday at the earliest. I need to run `php-cs-fixer` as well (and update the config probably).

glensc:
there's now example in readme how to use github pages:
- https://github.com/composer/satis#examples

ps: this PR needs rebase for merge

alcohol:
@mikeschinkel I suspect you no longer need this functionality (possibly even moved on to a new employer by now). However, if you still have the time to rebase this PR and quickly test if it works as expected, I'll merge it for you. If not, I'll close this PR in a while.

PR #449, opened 2017-12-21
[CS] Remove unused imports

carusogabriel:
I've removed some unused imports. This save us some lines :put_litter_in_its_place: 

Should we enable it in `php_cs`?


alcohol:
I think it was disabled on purpose, but I cannot recall why.

PR #450, opened 2017-12-21
[CS] Update and run PHPCs

carusogabriel:
I've updated usage of deprecated rules and also ran `PHP-CS-Fixer` :wrench:

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,5 +1,7 @@
 <?php
 
+/* This config requires PHP-CS-Fixer version ^2.9 */
+
 $header = <<<EOF
 This file is part of composer/satis.
 
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -24,8 +24,8 @@ return (new PhpCsFixer\Config('satis'))
         '@PSR2' => true,
         '@Symfony' => true,
         // additionally
-        'concat_with_spaces' => true,
-        'concat_without_spaces' => false,
+        'array_syntax' => ['syntax' => 'short'],
+        'concat_space' => false,
         'header_comment' => ['header' => $header],
         'no_unused_imports' => false,
         'no_useless_else' => true,
@@ -33,9 +33,8 @@ return (new PhpCsFixer\Config('satis'))
         'ordered_imports' => true,
         'phpdoc_align' => false,
         'phpdoc_order' => true,
-        'phpdoc_params' => false,
+        'phpdoc_align' => false,
         'phpdoc_summary' => false,
-        'short_array_syntax' => true,
         'simplified_null_return' => false,
     ])
     ->setFinder($finder)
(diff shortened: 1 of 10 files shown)


src/Builder/ArchiveBuilder.php
```diff
@@ -113,11 +113,11 @@ public function dump(array $packages)
                     ;
 
                     if (!file_exists($path)) {
-                        $downloadDir = sys_get_temp_dir() . '/composer_archiver/' . $packageName;
+                        $downloadDir = sys_get_temp_dir().'/composer_archiver/'.$packageName;
```
sokolnikov911: I think this does not correspond to psr-2 [https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide-additions.md#strings-and-concatenation](url)

carusogabriel: I guess I used the wrong rule. Gonna change it

alcohol:
Maybe add a docblock comment in the `.php_cs.dist` file to specify which version of php-cs-fixer is required (e.g. in the form of a semver constraint as supported by Composer, e.g. `^2.0` or higher).

carusogabriel:
@alcohol Let just see if I understood: you think we should add something like:
```php
/** 
 * @version ^2.0
 */
```
or no, a rule from `PHP-CS-Fixer` to specify the version?

alcohol:
That is what I suggested, yes :-)

.php_cs.dist
```diff
@@ -17,25 +17,27 @@ $finder = (new PhpCsFixer\Finder())
     ->in(__DIR__.'/views')
 ;
 
+/**
+ * @version ^2.0
```
carusogabriel: @alcohol Is this valid? :smile:

alcohol: I would not use @version since that is sort of "official".

Can I propose:

```
/* This config requires php-cs-fixer version ^2.3 */
```

alcohol: Or whichever version we require :-)

carusogabriel: As we support `PHP 7`, we can support PHPCS version `^2.9` :smile:

alcohol:
Thanks

denmasgeoo:
php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
php -r "if (hash_file('SHA384', 'composer-setup.php') === '544e09ee996cdf60ece3804abc52599c22b1f40f4323403c44d44fdfdd586475ca9813a858088ffbc1f233e9b180f061') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
php composer-setup.php
php -r "unlink('composer-setup.php');"

PR #451, opened 2017-12-29
added a "_comment" field into satis schema

staabm:
support a `"_comment"` field to describe "usefull" things in the statis.json file.

this is kind of a feature parity change, as this is also supported in composer.json

https://github.com/composer/composer/blob/57b3e2451404af0dac0f5fd797fce6789fa2624b/res/composer-schema.json#L513

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -252,6 +252,10 @@
         "notify-batch": {
             "type": "string",
             "description": "a URL that will be called every time a user installs a package."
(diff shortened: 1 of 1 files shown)


staabm:
my use case: I want to link satis related docs for my future me, as I think the satis setup will only need very few changes and therefore having something at the hand would be great.

alcohol:
Cheers

PR #455, opened 2018-01-16
added branch-alias to a version's title

staabm:
since branch aliases are rather important to us it would be very helpful when those would show up in the satis index.html page.

the PR adds this information to the `title` attribute of each version, like shown on the screen below:

grafik


staabm:
Thx for the fast merge!

PR #460, opened 2018-02-22
Output the (pretty) version if the progress bar is disabled

MatthiasKuehneEllerhold:
I want to see which version of the package the build-command is currently dumping.
So I've added it.


alcohol:
Hmm, slightly odd phrasing, but I can adjust that later. Thanks :-)

MatthiasKuehneEllerhold:
Thanks for merging, my english is not the yellow from the egg! ;-)

PR #463, opened 2018-03-01
Avoid calling the length Twig filter on null

mlocati:
This fixes the following `Twig_Error_Runtime` error thrown when using PHP 7.2:

```
count(): Parameter must be an array or an object that implements Countable
```


alcohol:
Thanks 👍

PR #465, opened 2018-03-06
added a mouseover title on "last updated" text

staabm:
instead of a estimate like "a few minutes ago" or similar one can get a exact timestamp when hovering the text in the website

grafik


alcohol:
Would prefer to see DATE_COOKIE used instead.

staabm:
here we go.

alcohol:
Thanks :-)

PR #468, opened 2018-03-08
capture blackfire metrics for subprocesses

staabm:
after my support question to blackfire with 
https://twitter.com/markusstaab/status/970587335778762752

I got a great answer of the support to utilize [metrics argument capturing](https://blackfire.io/docs/reference-guide/metrics#metrics-capturing-arguments) to get a better understanding of which cli commands are actually the problem (as satis executes 98% of its time within subprocess commands and until now noone could say which are actually the slow ones).

so with this change one can get a better picture of what is going on when running blackfire profiles. 

before this patch
grafik

with this patch
grafik

### DIFF
--- a/.blackfire.yml
+++ b/.blackfire.yml
@@ -1,0 +1,8 @@
+metrics:
+  composer.processes: # metric name
+    label: "composer subprocesses"
+    matching_calls:
+      php:
+        - callee:
+            selector: '=Composer\Util\ProcessExecutor::execute' # aggregate the costs of all Cache::write calls
+            argument: { 1: "^" }  # but create separate nodes by the first argument value
--- a/.blackfire.yml
+++ b/.blackfire.yml
@@ -4,5 +4,5 @@ metrics:
     matching_calls:
       php:
         - callee:
-            selector: '=Composer\Util\ProcessExecutor::execute' # aggregate the costs of all Cache::write calls
+            selector: '=Composer\Util\ProcessExecutor::execute' # aggregate the costs of all ProcessExecutor::execute calls
             argument: { 1: "^" }  # but create separate nodes by the first argument value


alcohol:
Is this ready to be merged? Can you elaborate a bit more on what the motivation/purpose behind this is?

staabm:
This is ready to be merged.

The main point of this change is to tell blackfire that it should seperate the aggregation of perf information per $command arg of the ProcessExecutor.
Essentially this leads to a profile in which expensive console commands are shown seperately.
Without this change blackfire will tell you that the ProcessExecutor takes a lot of time (as satis spends huge parts of its runs while running subprocesses) and you can see which exact callsite is the culprit.

So after all this is a DX change which tells the profiler some more things about this app so it can create better/more usefull performance profiles for us.

Does this make sense?

alcohol:
While I understand the use of this, I do question whether or not this belongs in this repository. It is not a part of this application. It is merely helpful for developers that are attempting to debug further details specifically through blackfire. Don't get me wrong, I think it's great you're investing time into this. But do you really feel like this should be committed into this repository? Would it not make more sense to perhaps simply document it? Just wondering how you feel about this :-)

staabm:
Its a small single file and having it in the repos gives immediate benefit everyone which uses blackfire (even if they dont know this feature etc).

Having it in the docs would be a alternative but I guess a lot of people would then miss it and have trouble profilling(or at least wont get the max. available features out of it).

I am fine with whatever solution you like

alcohol:
If you would be willing to take the time to add it to the documents, I would appreciate that very much (in this same PR). I'm OK with keeping the file as well, in that case. The purpose is clearly defined in that case in the documentation :-)

staabm:
an alternative for this PR would be to add something like composers `--profile` option.

alcohol:
It's ok.

PR #470, opened 2018-03-27
Added "enforce-dependencies" option to enforce constraints on packages

mikaelmeulle:
Context/Problem : when using satis with the option "require-dependencies" : "true" and not in "require-all" mode, satis is resolving required dependencies  to deduce which package versions should be downloaded. When you set a constraint on a package in the satis json, unfortunately if a dependency (or a dependency of a dependency) is declaring  broader versions, then satis will fetch those additionnal versions.

Proposed solution : i added a new option named "enforce-dependencies" which allow to restrict dependencies on downloaded packages. 

Remark : the patch is really not perfect, and i'm new to the satis source code, i will be happy to discuss that change.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -123,6 +123,10 @@ If true, selects all versions of all packages in all repositories defined.
 
 If true, resolve and add all dependencies of each required package.
 
+### enforce-dependencies
+
+If true, honor root requirements when resolving dependencies of required packages.
+
 ### require-dev-dependencies
 
 If true, resolve and add all development dependencies of each required package.
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -137,6 +137,10 @@
             "type": "boolean",
             "description": "If true, resolve and add all dependencies of each required package."
         },
+        "enforce-dependencies": {
+            "type": "boolean",
+            "description": "If true, honor root requirements when resolving dependencies of required packages"
+        },
         "require-dev-dependencies": {
             "type": "boolean",
             "description": "If true, resolve and add all Dev dependencies of each required package."
--- a/src/Console/Command/BuildCommand.php
+++ b/src/Console/Command/BuildCommand.php
@@ -73,6 +73,8 @@ The json config file accepts the following keys:
   required to mirror packagist for example, setting this
   to true will make satis automatically require all of your
   requirements' dependencies.
+- <info>"enforce-dependencies"</info>: if you set to true the option require-dependencies, 
+required dependencies will always honor root requirements.
 - <info>"require-dev-dependencies"</info>: works like require-dependencies
   but requires dev requirements rather than regular ones.
 - <info>"only-dependencies"</info>: only require dependencies - choose this if you want to build
--- a/src/PackageSelection/PackageSelection.php
+++ b/src/PackageSelection/PackageSelection.php
@@ -669,7 +669,7 @@ class PackageSelection
      *
      * @return Link[]
      */
-    private function selectLinks(Pool $pool, array $links, bool $isRoot, bool $verbose, ConstraintInterface[] $enforced_constraints): array
+    private function selectLinks(Pool $pool, array $links, bool $isRoot, bool $verbose, $enforced_constraints): array
     {
         $depsLinks = $isRoot ? [] : $links;
 
--- a/src/PackageSelection/PackageSelection.php
+++ b/src/PackageSelection/PackageSelection.php
@@ -690,7 +690,7 @@ class PackageSelection
             } elseif (is_a($link, Link::class)) {
                 $name = $link->getTarget();
                 $link_constraint = $link->getConstraint();
-                if ($this->enforceDependencies && array_key_exists($name, $enforced_constraints) && $link_constraint->getPrettyString() != $enforced_constraints[$name]->getPrettyString()) {
+                if ($this->enforceDependencies && $enforced_constraints && array_key_exists($name, $enforced_constraints) && $link_constraint->getPrettyString() != $enforced_constraints[$name]->getPrettyString()) {
                   $link_constraint = new MultiConstraint([$link_constraint,  $enforced_constraints[$name]], true);
                   $this->output->writeln('Enforce constraints on package '.$name. ': '. $link->getConstraint().' -> '.$link_constraint);
                 }
--- a/src/PackageSelection/PackageSelection.php
+++ b/src/PackageSelection/PackageSelection.php
@@ -29,6 +29,8 @@ use Composer\Repository\PlatformRepository;
 use Composer\Repository\RepositoryInterface;
 use Composer\Semver\Constraint\EmptyConstraint;
 use Composer\Semver\VersionParser;
+use Composer\Semver\Constraint\ConstraintInterface;
+use Composer\Semver\Constraint\MultiConstraint;
 use Composer\Util\Filesystem;
 use Symfony\Component\Console\Output\OutputInterface;
 
@@ -63,6 +65,9 @@ class PackageSelection
 
     /** @var bool Filter dependencies if true. */
     private $requireDependencyFilter;
+
+    /** @var bool always honor root requirements if true. */
+    private $enforceDependencies;
 
     /** @var string Minimum stability accepted for Packages in the list. */
     private $minimumStability;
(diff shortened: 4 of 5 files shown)


alcohol:
To clarify, the proposed changes are meant to make sure that the smallest set of matching versions is downloaded/cached, rather than the broadest set?

If so, I find the name chosen rather misleading.

Also, it would probably be better if such options (or filters) were to be extracted into their own method or class. They could then be applied to the full set of packages/versions that are initially resolved. 

Right now we just keep adding more and more functionality to the same method/class, thus greatly increasing the complexity of said method/class. This is not desirable.

mikaelmeulle:
hello, thanks for the reply.
The proposed change ensure that the version constraints set up in the top level satis.json file are not violated when satis downloads dependencies and dependencies of dependencies and so on... 
If you consider my proposal as a good feature, i can make the refactor requested and commit it.
thanks in advance
best

alcohol:
So basically as I understand it, if you have several constraint ranges for the same package, you want to limit the amount of versions to the smallest set (basically only the set that matches all overlapping constraint ranges), yes?

mikaelmeulle:
hello, this is almost the idea but not exactly. The code pushed does not limit the amount to the smallest set.
at least, it limits the versions of each packages to not violate version constraints set up in the satis json file. that was my basic idea.

i'm not sure if i am clear, let me know

alcohol:
I think you made a mistake while rebasing/merging. Your commits/changes don't really make sense anymore.

mikaelmeulle:
hello, ok thank you ! should i re-open it ?

alcohol:
If you can just fix your branch in your fork, that would be enough.

alcohol:
There are conflicts currently. Are you sure your local master branch was up to date?

mikaelmeulle:
hello, thank you for the answer, i will re-post it today, best

Mikael

simesy:
I *believe* that https://github.com/composer/satis/pull/549 closes this.

alcohol:
I'm not sure. It looks similar. Unfortunately I do not have the time to compare the code paths in depth.

alcohol:
@mikaelmeulle Can you please verify if the recent changes/additions cover your needs/requirements. And if so, can I close this PR? If not, can you update/rebase this PR, and possibly add that which is missing still?

mikaelmeulle:
hello,
yes i will do it and test on the week of July 22th.
thank you very much for following the issue :)

mikaelmeulle:
hello, 

i just did a rebase and simplification of my original PR..

basically what i was trying to achieve is still not done by the PR https://github.com/composer/satis/pull/549

the idea of the enforce-dependencies option is to honor root requirements constraints when resolving dependencies.

For example, in my root satis.json i require the package "drupal/core": "^8.7"  and another package "drupal/xxx".
the package drupal/xxx has a dependency to drupal/core which version is very broad, for example "drupal/core": "^8"

my patch will do "^8, ^8.7" when versions of drupal/core will be resolved.

I hope i am enough clear :)

best,

Mikael

alcohol:
Not sure I understand..

> my patch will do "^8, ^8.7" when versions of drupal/core will be resolved.

That seems pointless because ^8.7 exceeds ^8, so there is no point to having the first constraint.

mikaelmeulle:
hello,
i think i would need a concrete example to help people understand. Maybe my example was not clear enough or not well chosen. 

The aims of the patch is to always honor top-level constraints of packages listed.

do i need to build a new example?

mikaelmeulle:
due to the recent commits, i have to re create the patch anyway :)
i will post a new example with the new version of the patch, will make this in the next days.

simesy:
Now that we have `only-best-candidates` and `blacklist` I think we got the outcome we needed.

But i suspect this PR could serve the same purpose only more elegantly ... so I'll be free to review and test this PR on our current setup, and perhaps write a test for it.

simesy:
On second thoughts, I think it's slightly different. But I will still help with a review and test since I understand the goal.

mikaelmeulle:
hello,
yes the recent options are very good for Satis.
the enforce-dependencies is acting like a whitelist, but it can have an unexpected outcome because in some edge cases some dependencies will not be resolved if they are incompatible with the top-level requirements.(the same with blacklist i guess)  But it is the idea of enforce-dependencies at the end, to ensure only some versions are synchronized . We use it for more than 1 year and a half now in a broad organization (for composer-based Drupal 8 applications), and it work fine and achieve the whitelisting strategy we want.
i will post my code today, stay tuned :)

mikaelmeulle:
Here is my **satis.json** to see the great benefits of "enforce-dependencies" option, espacially in the Drupal ecosystem.

```
{
    "name": "test/test",
    "homepage": "http://x.x.x.x/",
    "config": {
    },
    "repositories": {
        "composer": {
            "type": "composer",
            "url": "https://packagist.org"
        },
        "drupal": {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    },
    "require": {
		"drupal/core": "^8.7.8",
        "drupal/address": "^1.4"
    },
	"require-all": false,
    "require-dependencies": true,
    "enforce-dependencies": false,
    "minimum-stability": "stable",
    "require-dependency-filter": true,
    "archive": {
        "directory": "dist",
        "format": "zip",
        "checksum": false,
        "skip-dev": true
    }
}
```

only two packages are in the root requirements :+1: 

- drupal/core
- drupal/address

in the package, drupal/address, composer.json looks like this : 
```
    "require": {
        "drupal/core": "^8.6",
        "commerceguys/addressing": "^1.0.0"
    },
``` 
and other dependencies of drupal/core and their dependencies, may have broader versions requirements on "drupal/core" 

**when "enforce-dependencies" is set to true** :  the only drupal/core version downloaded are "8.7.8" and above.

**when "enforce-dependencies" is set to false** :  drupal/core is downloaded in the following versions :  8.7.8, 8.6.0, 8.0.0

simesy:
Hiya! Thanks for the sample.
I tested with only-best-candidates and I only got 8.7.8 per screenshot.
Screen Shot 2019-10-23 at 9 48 34 am

Config I used:
```
{
    "name": "test/test",
    "homepage": "http://x.x.x.x/",
    "config": {
    },
    "repositories": {
        "composer": {
            "type": "composer",
            "url": "https://packagist.org"
        },
        "drupal": {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    },
    "require": {
        "drupal/core": "^8.7.8",
        "drupal/address": "^1.4"
    },
	"require-all": false,
    "require-dependencies": true,
    "only-best-candidates": true,
    "minimum-stability": "stable",
    "require-dependency-filter": true
}

```

mikaelmeulle:
hello,
thank you for your answers. In that specific example, "enforce-dependencies" and "only-best-candidates" do the same but it is only an accident :)

in the example, "enforce-dependencies" option ensure that when the package "drupal/core" is synchronized, the versions will ALWAYS match the root requirement (in my example "^8.7.8").
Therefore, if multiple versions match, multiple versions will be downloaded.

in the example, "only-best-candidates":   option ensure that when the package "drupal/core" is synchronized, only the bests candidates will be fetched for the package "drupal/core"  (i just looked at the code in composer repo doing the job). This diverge with enforce-dependencies in at least two points : "only-best-candidates" will restrict the number of versions fecthed, and most important it will not honor the root requiremen. If it honors the root requiremennt, it will be by chance...

The good thing is that "only-best-candidates" and "enforce-dependencies" may be used together  to get the best of both worlds :)

Let's take the new example bellow,
only-best-candidates will fetch      8.7.8, 8.7.7, 8.7.6, 8.7.5  versions of drupal/core
enforce-dependencies will fetch       8.7.7, 8.7.6, 8.7.5  versions of drupal/core :+1: 

```
{
    "name": "test/test",
    "homepage": "http://x.x.x.x/",
    "config": {
    },
    "repositories": {
        "composer": {
            "type": "composer",
            "url": "https://packagist.org"
        },
        "drupal": {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    },
    "require": {
        "drupal/core": "^8.7.5, < 8.7.8",
        "drupal/address": "^1.4"
    },
    "require-all": false,
    "require-dependencies": true,
    "enforce-dependencies": true,
    "only-best-candidates": false,
    "minimum-stability": "stable",
    "require-dependency-filter": true,
    "archive": {
        "directory": "dist",
        "format": "zip",
        "checksum": false,
        "skip-dev": true
    }
}
```


what do you think? is it more clear now ?

simesy:
Thanks for updating! I'll have another go.

simesy:
(For testing speed i'm removing the archive section, I don't think it would make a difference?)

mikaelmeulle:
hello,
did you find time to make some tests, what do you think after all :)?
thanks in advance, 
all the best

tobybellwood:
I've been trying to get this to work today, with no joy - it looks like the MultiConstraint at https://github.com/composer/satis/pull/470/files#diff-82f59585c3ee5337a338ebd586c3cba7R694 is malforming the string somehow (leaving the [] in)

```
Enforce constraints on package drupal/config_filter: [>= 1.0.0.0-dev < 2.0.0.0-dev] -> [[>= 1.0.0.0-dev < 2.0.0.0-dev] == 1.4.0.0]

In VersionParser.php line 477:
                                                                               
  [UnexpectedValueException]                                                   
  Could not parse version constraint < 2.0.0.0-dev]: Invalid version string "  
  2.0.0.0-dev]"    
```

simesy:
> only-best-candidates will fetch 8.7.8, 8.7.7, 8.7.6, 8.7.5 versions of drupal/core

Sadly`--only-best-candidates` has a deficiency, and in the [described example](https://github.com/composer/satis/pull/470#issuecomment-545356124) above it should only get one version (ie. it should get only one "best candidate" per root requirements).

Otherwise the two features are similar in that they require some context of the top level object to make some decisions. The solution in this patch is to pass this `$enforced_constraints` parameter into the process, which boils down to "I need to make a logical decision during the resolution process, and I need access to some top-level information to do that."

I've looked at something similar for the `--only-best-candidates` problem that I have but it feels like it is coercing satis to do something fundamentally different to what it's designed for. 

Sorry I don't think I'm adding a lot of value added here but trying to document what I saw.

PR #480, opened 2018-07-14
Fix the error when ther is no downloaded archive.

zerustech:
There is an issue in commit 73fcd34dbaffe11e3e6f70bc7975d1c3f41630fc:
when there is no downloaded archive, changing the `dist-type` of current
package makes the `download manager` fail to download the archive from
the remote dist url, so it has to download the archive from the source url.

This commit fixes the issue above by checking if the correct archive exists or not. If not, it restores the `dist-type` of current package, and let the `download manager` to download the archive from remote dist url. After the archive has been downloaded, the archive will be renamed according to the archive format.


zerustech:
@alcohol Thanks!

alcohol:
Thank you too

PR #488, opened 2018-08-07
Streamline build

ph-fritsche:
**These changes enable the use of the resulting composer-repository as the single source for installing one or multiple packages defined by the primary requirements while heavily reducing the required space and time for building.**

#### Add config option "repositories-dep" with repositories only used for dependencies

Enables the use of (composer-)repositories like packagist.org in combination with _require-all_.
This makes it possible to have everything from the "primary" repositories and all dependencies in one place. No extra tools for automatic rewriting of _require_ needed.

#### Reduce packages included per dependencies to needed versions

Satis used to include all versions of dependencies into the selection.

With this change it will still select all versions of the primary/root packages defined by _require_ or _require-all_, but reduces the selection of dependencies to those versions which would be selected by Composer when installing one of the primary packages (i.e. possible combinations of _prefer-lowest_ and _prefer-stable_).

Since Composer never installs dev-dependencies for dependencies these are also removed from the selection

#### Add command line option "repository-strict" and provide consistent behavior of repository-filter and packages-filter if it is not set

When running build with _packages_ arguments, Satis would reduce the primary requirements to those packages, but still would install all dependencies.
When running build with _repository-url_ option, Satis would reduce the primary requirements to those it can find in this repository and then install those dependencies it can also find in that repository.

I still have to find the reason, why a combination of both is explicitly made impossible, although it might be desired if you know, that you made changes in only a specific subset of the primary requirements.
But for now the rewritten selection provides consistent behavior resulting in all available dependencies present in the result - no matter how they were found.

#### Add config option "archive"."rearchive"

For dumping an archive Satis used to download and extract packages like Composer would and then create an archive out of the resulting directory.
If _archive.rearchive_ is set to false and a package already provides a dist of type tar or zip, Satis will now just copy that file.

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -80,6 +80,10 @@
                 "override-dist-type": {
                     "type": "boolean",
                     "description": "If true, archive format will be used to substitute dist type when generating archive file name."
+                },
+                "rearchive": {
+                    "type": "boolean",
+                    "description": "Create new archives for packages with a tar or zip dist, defaults to true"
                 }
             }
         },
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -88,6 +88,11 @@
             "description": "A set of additional repositories where packages can be found.",
             "additionalProperties": true
         },
+        "repositories-dep": {
+            "type": ["object", "array"],
+            "description": "A set of additional repositories where packages for dependencies can be found.",
+            "additionalProperties": true
+        },
         "minimum-stability": {
             "type": "string",
             "description": "The minimum stability the packages must have to be install-able. Possible values are: dev, alpha, beta, RC, stable.",
--- a/src/Builder/ArchiveBuilder.php
+++ b/src/Builder/ArchiveBuilder.php
@@ -13,6 +13,9 @@ namespace Composer\Satis\Builder;
 
 use Composer\Composer;
 use Composer\Factory;
+use Composer\Downloader\DownloadManager;
+use Composer\Package\PackageInterface;
+use Composer\Package\Archiver\ArchiveManager;
 use Composer\Util\Filesystem;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
@@ -39,11 +42,8 @@ class ArchiveBuilder extends Builder
         $helper = new ArchiveBuilderHelper($this->output, $this->config['archive']);
         $basedir = $helper->getDirectory($this->outputDir);
         $this->output->writeln(sprintf("<info>Creating local downloads in '%s'</info>", $basedir));
-        $format = $this->config['archive']['format'] ?? 'zip';
         $endpoint = $this->config['archive']['prefix-url'] ?? $this->config['homepage'];
         $includeArchiveChecksum = (bool) ($this->config['archive']['checksum'] ?? true);
-        $ignoreFilters = (bool) ($this->config['archive']['ignore-filters'] ?? false);
-        $overrideDistType = (bool) ($this->config['archive']['override-dist-type'] ?? false);
         $composerConfig = $this->composer->getConfig();
         $factory = new Factory();
         /* @var \Composer\Downloader\DownloadManager $downloadManager */
@@ -133,23 +133,9 @@ class ArchiveBuilder extends Builder
                     $archiveFormat = 'file';
                 } else {
                     $targetDir = sprintf('%s/%s', $basedir, $intermediatePath);
-                    $archiveFormat = $format;
 
-                    if (true === $overrideDistType) {
-                        $filesystem = new Filesystem();
-                        $filesystem->ensureDirectoryExists($targetDir);
-                        $originalDistType = $package->getDistType();
-                        $package->setDistType($format);
-                        $path = realpath($targetDir) . '/' . $archiveManager->getPackageFilename($package) . '.' . $format;
-
-                        if (!file_exists($path)) {
-                            $package->setDistType($originalDistType);
-                            $downloaded = $archiveManager->archive($package, $format, $targetDir, null, $ignoreFilters);
-                            $filesystem->rename($downloaded, $path);
-                        }
-                    } else {
-                        $path = $archiveManager->archive($package, $format, $targetDir, null, $ignoreFilters);
-                    }
+                    $path = $this->archive($downloadManager, $archiveManager, $package, $targetDir);
+                    $archiveFormat = pathinfo($path, PATHINFO_EXTENSION);
                 }
 
                 $archive = basename($path);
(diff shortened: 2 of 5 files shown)


ph-fritsche:
Let me add some data...

Given a local git repo with just one commit:
```
{
    "name": "ph-fritsche/satis-test-source",
    "require":{
        "composer/composer": ">=1.6"
    }
}
```
And a satis.json with:
```
{
    "name": "test repo",
    "homepage": "http://127.0.0.1:8001",
    "repositories": [
        {
            "type": "vcs",
            "url": "file://./test-source"
        },
        {
            "type": "composer",
            "url": "https://packagist.org"
        }
    ],
    "require": {
        "ph-fritsche/satis-test-source": "*"
    },
    "require-dependencies": true,
    "output-dir": "srv",
    "providers": true,
    "archive": {
        "directory": "dist",
        "format": "zip",
        "skip-dev": false
    }
}
```
Running the satis build (each time after wiping the directory and clearing the cache) cost on that machine:
```
at c5fddd28	58 min	134 MB
at f7b98563	3:20 min	5.5 MB
at f7b98563 with _archive.rearchive_=false	1:23 min	5 MB
```
Each with the same result:
That when serving that repository ph-fritsche/satis-test-source can be installed completely via
```
composer create-project --repository http://127.0.0.1:8001 --no-secure-http -s dev ph-fritsche/satis-test-source
```

Automating such a build and making it available for a staging area makes it possible to stage and deploy the primary packages fast and reliable using only one repository.

And by using _require-all_ in combination with _repositories-dep_ it is also possible to connect multiple repositories and add their dependencies from other resources without maintaining _require_.
This way it is even possible to chain multiple Satis-repositories and maintain connection to a vcs or artifact only once anywhere in the chain.

I think that makes Satis a very powerful tool for "just a small file based repository" ;)

alcohol:
My main concern/question is, if I merge this, will it affect existing installations of Satis?

E.g. will it change their behaviour if users keep using the same configuration they already had? And could this lead to unexpected or undesired results?

ph-fritsche:
Yes, it could affect existing installations.
No, it will not affect existing installations if they don't rely on undesired side-effects of the previous implementation or reflect very rare cases.

Everything defined by _repositories_ and _require_/_require-all_ will be available like before.
Everything included via _require-dependencies_ or _require-dev-dependencies_ will be available like used by Composer's install or update.

Given an existing installation defines packageA as requirement,
packageA requires packageB >=1 ,
and packageB exists as 1, 2, 3, 4 .
Composer only will use 1 or 4 depending on _prefer-lowest_. 2 and 3 are not part of the repository.
If the installation relies on packageB being available as 2, it will break.

Or given an existing installation defines packageA and packageB as requirements,
packageA requires packageC 1 || 3 || 5 ,
and packageB requires packageC 2 || 3 || 4 .
The intersect packageC@3 is not part of the repository.
If the installation relies on packageA and packageB being available together, it will break.

But note that this is only the case if those requirements are included as dependencies and are not part of _require_ in satis.json .
I guess everyone who tried to build such a repo, waited for a few hours (or days) and switched to maintaining _require_ (manually or e.g. by composer-satis-builder) - or gave up completely and included dependencies from another repository.

alcohol:
I am a bit confused by your explanation of how dependencies 1 2 3 and 4 would only have 1 and 4 available depending on prefer-lowest. Did you take into account that platform requirements could mean that perhaps 4 would not be applicable and 3 could be the highest some target hosts can support? All of this cannot really be predicted up front, so I am a bit sceptical about this change in particular.

ph-fritsche:
Yes, this would lead to the target host using 1 instead of 3 because 3 is filtered out by the repository and 4 is filtered out by the target host.
There are scenarios where using that repository as the single source might break the dependency resolving on a target host.

My point is: Using Satis as single source for installations without managing requirements beforehand was not a practical solution before because of the costs - and it is very unlikely that anyone actually has such an installation in place.

If you included 3 into _require_ it is not filtered.
And because you can now chain repos with _require-all_, future users knowing of such conflicts on their targets will only have to add that potentially live-saving package version 3 into their initial requirements and subsequent mergers will carry it through.

But I see your point and will make the reduction optional.
Do you really think that backward-compatibility is such a big issue here? Otherwise i'd very much like to change the default behavior towards the faster and smaller solution instead of letting everyone run into that monster build process before searching for other options.

alcohol:
To be honest, I am not sure. The Satis user community is relatively small. Could be these changes hurt some setups our there, could be nobody notices a thing. I will merge it, and we will have this thread to fall back on in case anyone asks questions or runs into unexpected issues :-)

alcohol:
I find the `repositories-dep` name a bit awkward btw. In general, configuration options should never be abbreviated. Their intent should be as obvious and clear as possible from the name itself.

ph-fritsche:
Yep. If you combine big repositories with _require-all_ you end up waiting forever.
Leading to no possible way to say, "I want everything in my sources and everything these packages need from packagist/this big other repo."

SvenRtbg:
I'm following this discussion and I am a little bit worried about that efficiency change filtering out versions.

My setup is as follows: The first satis run requires all kinds of dependencies (which are either required by `*` or without an upper limit like `>=2.15.3`). It creates a repo with the selected packages and their dependencies, and the selection usually is including way more versions that I'd ever need - including the ones that are below my requirement, but used because that package is required somewhere else in a broader version range. The second step `require-all` my local repositories and the satis repo from step 1. It compiles ZIP files of everything, and the resulting satis repo is used in every internal software as the only source for dependencies.

The definition of my requirements for external software until now was "fire-and-forget" - every new version was drawn into the repository, and all existing versions were kept.

I cannot have versions disappear, neither from internal nor external packages, just because Satis would be able to create a definition gap just by having new versions released (i.e. I require `>=1`, and over time v1, v2 and v3 are released, and v2 vanishes the very moment v3 is detected). 

I'd fear I'd have to constantly monitor version releases and adjust my version requirements - a maintenance task that wasn't necessary before.

alcohol:
As far as I understood it, it will still include all versions for root requirements?

The new filtering merely applies to transitive dependencies.

ph-fritsche:
If you have Satis1 with _require_ packageA `>= 2.15.3`, all versions of packageA matching that constraint will be included.
If you include Satis1 in Satis2 with _require-all_, everything in Satis1 will be included, so all versions of packageA `>= 2.15.3` will be present.

And now comes the scenario, where I'm not sure if I understood you correctly; so please elaborate if i got something wrong:
If packageA@2.15.3 required packageB@`>=1` and you use Satis1 with _require-dependencies_, Satis would previously include all available versions of packageB@`>=1` in the repository.
But when you install packageA@2.15.3 via Composer and B@1, B@2, B@3 are present only B@1 or B@3 will be used.
So yes - the moment Satis1 finds B@3, B@2 will not be part of the Satis1 and subsequently not be part of Satis2 any more.

If B@3 breaks your build you will not get around defining a packageA@2.15.4 which requires packageB@`>=1 <3`.
Then you would end up with a repo with A@2.15.3 and A@2.15.4 as well as B@1 (via A@2.15.3 and A@2.15.4), B@2 (via A@2.15.4) and B@3 (via A@2.15.3).

You will never get around monitoring version releases when you build software with open-ended requirements.
The difference with this commit is: Now there is a clear distinction between what is part of the repo because you _required_ it to be, and what is required because it matches one of the packages dependencies.

SvenRtbg:
A simple example: A common dependency requirement for PHPUnit might look like this:
`"phpunit/phpunit": "^4.8.35 || ^5.7 || ^6.5 || ^7.1"`

As we know from the docs, each major version has a different minimum requirement for the PHP version it runs on. 4.x and 3.x run with every PHP `>=5.3.3`, 5.x requires `^5.6||^7.0`, 6.x requires `^7.0` and 7.x requires `^7.1`.

So if I don't require PHPUnit directly (which is only the example here), I understand that I'd end up including either `^4.8.35` or `^7.1`? If I am running PHP below 7.1, I'd be forced to use PHPUnit 4.8.

There is another pull request here that adresses package selection: #470 

From the description there I understand that it tries to enforce MY selection of versions more. From above example, if I require PHPUnit `>=6` in Satis and a package with above requirements, I'd expect the repository to NOT include any PHPUnit versions below 6 with that patch applied.

I try to find out if and how your change affects me, and how I can profit from it - therefore thanks for discussing it with me a bit. And the feature still lacks documentation, which might result from this thread as well.

ph-fritsche:
Yes, if your primary requirements depend on `"phpunit/phpunit": "^4.8.35 || ^5.7 || ^6.5 || ^7.1"` your repository only holds PHPUnit `^4.8.35` and `^7.1` and if you install that primary requirement with PHP `<7.1` you will run PHPUnit `^4.8.35`.
But according to the dependency-definition of your primary requirements that is just fine.

So if at some point you think phpunit `^4.8.35` is not acceptable any more. Just edit the requirements of your primary packages and the installation will take care of it the next time you run build.

I read the pull request you mentioned before writing my changes. From my understanding it does not solve the problem, because it introduces a manually maintained list of constraints enforced on the automatic generated list of dependencies.
So the chances are good that a package introduces new requirements and you include a bunch of packages you did not apply constraints to before - resulting in your build time skyrocketing.
And you would have to maintain such a list in the first place - e.g. PHPUnit `>=6` introduces a lot of new dependencies bringing up the same problem.

PR #490, opened 2018-08-11
Streamline build II

ph-fritsche:
### Add config option require-dependency-filter

When set to false, all versions matching a dependency are included.
See discussion here: #488 

### Add config option strip-hosts

When using Satis with sources that are only locally available, that repo carries ambiguous data.

With this option you can define a `List of domains, IPs, CIDR notations, '/local' (=localnet and other reserved) or '/private' (=private IPs) to be stripped from the output. If set and non-false, local file paths are removed too.`

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -126,6 +126,13 @@
             "type": "boolean",
             "description": "If false, will include all versions matching a dependency."
         },
+        "strip-hosts": {
+            "type": ["array", "boolean"],
+            "description": "List of domains, IPs, CIDR notations, '/local' (=localnet and other reserved) or '/private' (=private IPs) to be stripped from the output. If set and non-false, local file paths are removed too.",
+            "items": {
+                "type": "string"
+            }
+        },
         "include-filename": {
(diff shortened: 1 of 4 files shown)


alcohol:
Maybe `require-dependency-filter` should default to `false` btw, so the recently introduced change doesn't catch anyone unaware?

ph-fritsche:
> Do you really think that backward-compatibility is such a big issue here? Otherwise i'd very much like to change the default behavior towards the faster and smaller solution instead of letting everyone run into that monster build process before searching for other options.

If you want me to, i'll change it.
But for someone creating a new repo i guess it is much better to be done after a few minutes - and if he represents one of those discussed rare cases, searching for how to include more.
If the profit was smaller, i'd prefer backward-compatibility, but that option is game-changing when setting up Satis for the first time.

Your call. :stuck_out_tongue:

alcohol:
Leave it as-is for now. I'll try and keep an eye on the issues that we receive on this repository, and see if they are related or not :-)

PR #495, opened 2018-09-05
Fix #493

### LINKED ISSUE #493: Exception on build with repository filter
satis throws exception when running `build` with option `--repository-url` and either `require-dependencies` or `require-dev-dependencies` enabled.

```
PHP Warning:  Uncaught ErrorException: Object of class Composer\Repository\VcsRepository could not be converted to string in .../vendor/composer/satis/src/PackageSelection/PackageSelection.php:210
```

ph-fritsche:
Replace array_diff with array_filter


alcohol:
Thanks

PR #498, opened 2018-09-09
docker: use multistage build for docker build cache

glensc:
the change also includes minor Image size also decrease by 3MB due final runtime not having copy of composer:

```
$ docker images
REPOSITORY        TAG       IMAGE ID            CREATED              SIZE
composer/satis    latest    f5f0b17b376b        6 days ago           167MB
composer/satis    pr-498    2f5f8783d827        32 seconds ago       164MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
-FROM php:7-cli-alpine
+FROM composer:latest AS build
 
+COPY . /satis
+WORKDIR /satis
+RUN composer install --no-scripts --no-plugins --no-dev -a
+
+FROM php:7-cli-alpine
 MAINTAINER https://github.com/composer/satis
 
 RUN apk --no-cache add curl git subversion mercurial openssh openssl tini \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -20,18 +25,9 @@ ENV COMPOSER_ALLOW_SUPERUSER 1
 
 COPY php-cli.ini /usr/local/etc/php/
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
```
alcohol: I reckon we no longer need to copy composer into the final satis image then anymore, don't you think so too?

glensc: indeed. it's not needed there. removing

glensc:
tested build with:

```
$ docker build -t composer/satis .
$ git clone https://gist.github.com/glensc/e4e1fac01b331d11ecafb9e1878aa941 build
$ docker run --rm -it -v `pwd`/build:/build -v "${COMPOSER_HOME:-$HOME/.composer}:/composer" composer/satis
```

alcohol:
Sorry man, this PR must have gotten lost in my notifications. Thanks!

PR #506, opened 2018-10-01
dont render repo-config in initial view

staabm:
closes #504 

*BEFORE*

grafik

*AFTER*
- initial state

grafik

- after click on `Add this Satis repository to your composer.json`

grafik

PR #512, opened 2018-10-21
Update Satis UI to Bootstrap 4

antoniocambados:
This PR freshens up the Satis UI by upgrading it from Bootstrap 3 to Bootstrap 4. It also updates the JavaScript code, which has been converted to modern ES6 classes.

Some things worth noting:

- Both CSS and JS have been compiled using Webpack (with Webpack Encore for an easier setup).
- Overall assets size has dropped from ~280Kb to ~180Kb.
- Contains a longer explanation on how to set up the repository, which I largely took from [here](https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md#usage).

Here's how it looks:
Screenshot 1

And here is the detail of the setup info:
screenshot 2


alcohol:
I think the `/build/` output should be added to `.gitignore`, should it not? We can compile it in Travis before creating a container.

alcohol:
I think the `/build/` output should be added to `.gitignore`, should it not? We can compile it in Travis before creating a container.

alcohol:
I think the `/build/` output should be added to `.gitignore`, should it not? We can compile it in Travis before creating a container.

staabm:
I really like updating the UI. 

IMO you should make sure that the page header does not occupy that much space.
we recently updated the UI to use less space, see e.g.
https://github.com/composer/satis/pull/506
https://github.com/composer/satis/pull/505

staabm:
I really like updating the UI. 

IMO you should make sure that the page header does not occupy that much space.
we recently updated the UI to use less space, see e.g.
https://github.com/composer/satis/pull/506
https://github.com/composer/satis/pull/505

staabm:
I really like updating the UI. 

IMO you should make sure that the page header does not occupy that much space.
we recently updated the UI to use less space, see e.g.
https://github.com/composer/satis/pull/506
https://github.com/composer/satis/pull/505

alcohol:
I don't think the `/build/` output should be committed? Maybe add it to the `.gitignore`? We can compile it and bake it into the image during the Travis build.

staabm:
I really like updating the UI. 

IMO you should make sure that the page header does not occupy that much space.
we recently updated the UI to use less space, see e.g.
https://github.com/composer/satis/pull/506
https://github.com/composer/satis/pull/505

staabm:
I really like updating the UI. 

IMO you should make sure that the page header does not occupy that much space.
we recently updated the UI to use less space, see e.g.
https://github.com/composer/satis/pull/506
https://github.com/composer/satis/pull/505

alcohol:
I don't think the `/build/` output should be committed? Maybe add it to the `.gitignore`? We can compile it and bake it into the image during the Travis build.

alcohol:
Removed some duplicate comments as a result of the major outage Github had with their storage backend.

stof:
@alcohol this would hurt people installing Satis through composer, as they would not have the compiled assets

package.json
```diff
@@ -0,0 +1,19 @@
...
+  "devDependencies": {
+    "@symfony/webpack-encore": "^0.20.0",
+    "bootstrap": "^4.1.3",
+    "date-fns": "^1.29.0",
+    "jquery": "^3.3.1",
+    "node-sass": "^4.9.4",
+    "popper.js": "^1.14.3",
+    "sass-loader": "^7.1.0"
+  },
+  "license": "UNLICENSED",
```
stof: should be `MIT` to match the Satis license

views/assets/js/App/DateDistance.js
```diff
@@ -0,0 +1,19 @@
+import $ from 'jquery';
+import distanceInWordsToNow from 'date-fns/distance_in_words_to_now';
+
+class DateDistance {
+    static calculate($elements) {
```
stof: I would name it `$elements` only if it is already a jQuery collection. And in such case, you don't need to wrap it again in jQuery.

webpack.config.js
```diff
@@ -0,0 +1,48 @@
+var Encore = require('@symfony/webpack-encore');
+
+Encore
+    // directory where compiled assets will be stored
+    .setOutputPath('views/build/')
+    // public path used by the web server to access the output path
+    .setPublicPath('/assets')
```
stof: this is wrong. the public path is `build`, not `assets` (as you don't use the manifest, you are less impacted by the wrong config of course)

alcohol:
But should that not be part of their setup process (README should be updated)?

I mean, just a `git clone` is not sufficient for them either. They would still need to run `composer install`.

stof:
Btw, if you want to reduce the file size of assets, an option might be to drop jQuery too.

alcohol:
It would be nice to drop the jQuery dependency. Considering what little we use it for we could probably write that in vanilla JS just as well.

alcohol:
If you're feeling generous/creative/productive/brave/optimistic, you could maybe also take a stab at https://github.com/composer/satis/issues/513

stof:
@alcohol in which image ? For people installing Satis through composer, they download the github source, not an image.

webpack.config.js
cfv1000: The compiled files should be relative to the the "output" option from the cli command.

views/assets/js/App/DateDistance.js
antoniocambados: Dependency of jQuery for class DateDistance was removed.

webpack.config.js
antoniocambados: @codecov-io the current WebBuilder embeds every asset in the html code (css, js, images), so it's not possible for those to be relative to the "output-dir" option of the build command.

One thing I had in mind was making the build command to copy the compiled assets (or maybe symlink them if it's possible) to the "output-dir" and referencing them rather than embedding them. If you all want, I can give it a go and add this to the PR as well.

antoniocambados:
I think some messages have gone missing or misplaced. What a messy day on GitHub!

Regarding the compiled assets I think those should be included. If not, we would be forcing to compile them at least once. That could be done with a composer post-install script, but it would only succeed if the user has the required setup (node, npm, yarn...). And we're talking about a mandatory step here even if whoever installed this package does not intend to modify the UI whatsoever. Moreover, one could argue that the current version relies on and is including the compiled assets anyways :sweat_smile:

antoniocambados:
I think some messages have gone missing or misplaced. What a messy day on GitHub!

Regarding the compiled assets I think those should be included. If not, we would be forcing to compile them at least once. That could be done with a composer post-install script, but it would only succeed if the user has the required setup (node, npm, yarn...). And we're talking about a mandatory step here even if whoever installed this package does not intend to modify the UI whatsoever. Moreover, one could argue that the current version relies on and is including the compiled assets anyways :sweat_smile:

antoniocambados:
> It would be nice to drop the jQuery dependency. Considering what little we use it for we could probably write that in vanilla JS just as well.

Well that could certainly be done, but maybe replacing jQuery with a lightweight alternative such as [cash](https://github.com/kenwheeler/cash) is not a bad idea either. Overall assets size would be ~100Kbs

alcohol:
I am honestly not familiar with the current trend of javascript frameworks (both small and large) as I rarely do frontend development, so I'll defer to your judgement.

It also sounds like committing the compiled assets makes more sense, so keep it like that. Would you like to add any follow up work to this same PR? Or shall I go ahead and merge and then you can submit any other work in a new PR (or multiple even)?

package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "devDependencies": {
+    "@symfony/webpack-encore": "^0.20.0",
+    "bootstrap": "^4.1.3",
+    "date-fns": "^1.29.0",
+    "jquery": "^3.3.1",
```
staabm: would it make sense to use `jquery-slim` in this spot (as we used the slim variant also before this PR)?

or do some parts of bootstrap require the "full build"?

staabm:
> It also sounds like committing the compiled assets makes more sense, so keep it like that

I agree with it. in the past I just cloned satis from github and was done with it. no build etc. required.

staabm:
> I really like updating the UI.
> 
> IMO you should make sure that the page header does not occupy that much space.
> we recently updated the UI to use less space, see e.g.
> #506
> #505

any feedback regarding my above comment?

staabm:
> Well that could certainly be done, but maybe replacing jQuery with a lightweight alternative such as [cash](https://github.com/kenwheeler/cash) is not a bad idea either. Overall assets size would be ~100Kbs

in the past, the satis website didn't have any external depedencies. everything was embedded in the html document. not sure why, but maybe this was a "feature" which we need @alcohol ?

alcohol:
> in the past, the satis website didn't have any external depedencies. everything was embedded in the html document. not sure why, but maybe this was a "feature" which we need @alcohol ?

I think it was mostly a matter of simplicity. It is not a requirement as far as I know.

package.json
antoniocambados: The `jquery-slim` version is not offered as an npm package. The `jquery` package contains a compiled version of `jquery-slim` though, but funnily, if you require that instead of the regular `jquery`, the bundled js is significantly larger.

Bootstrap 4 requires `jquery` in order to run it's javascript features, in addition to `popper` if you want to use popups, tooltips and the like.

As of now, the PackageFilter class (which I ported directly from the previous code) makes heavy use of `jquery`, and I've also introduced the use of Bootstrap's collapsible plugin.

I've been working a bit on this but I think maybe it's better to be done in another PR once this is merged.

antoniocambados:
I think enabling external dependencies would make it easier to improve the web UI. I'll be doing another PR with this soon.

antoniocambados:
> > I really like updating the UI.
> > IMO you should make sure that the page header does not occupy that much space.
> > we recently updated the UI to use less space, see e.g.
> > #506
> > #505
> 
> any feedback regarding my above comment?

I can work a bit on that :smiley:

antoniocambados:
> I am honestly not familiar with the current trend of javascript frameworks (both small and large) as I rarely do frontend development, so I'll defer to your judgement.
> 
> It also sounds like committing the compiled assets makes more sense, so keep it like that. Would you like to add any follow up work to this same PR? Or shall I go ahead and merge and then you can submit any other work in a new PR (or multiple even)?

Before merging, as @stof pointed out I'll be fixing the manifest and updating the docs for frontend contribution instructions, plus I'll try to make the header more compact too as I posted above.

SvenRtbg:
I'd prefer not to link to external resources if possible. That way the page always fully works, even in internal networks with limited remote connections.

staabm:
> I'd prefer not to link to external resources if possible. That way the page always fully works, even in internal networks with limited remote connections.

to be more precise: I meant external js/css files on the same local host, not remote css/js files.

the current version of satis embedds everything into the html document.
I dont want it to request resources which are only available from a host which is connected to the internet.

alcohol:
None of these resources are external. They are all local to the project. The only change is that the javascript is no longer completely embedded within the html, but instead served separately.

antoniocambados:
Fixed the configuration for the (unused) `manifest.json` and brought to you a new header I'm really happy with: sleek, more compact yet with more info and useful links. Have a look:
satis-header-toggle

alcohol:
Nicely done. I'll merge this to keep the changeset relatively small. Any follow up work, please submit a new PR. I will happily review and merge those too :-)

staabm:
just pulled the latest master into our instance but didn't got the new design.

is a build step or similar required?

staabm:
ohh I had to hard-refresh the page in the browser.

antoniocambados:
@staabm Satis does not reference stylesheets, JS or icons, it embeds them in the HTML response. I was going to separate assets in this PR, but decided to go for that in a follow-up PR, focused on making satis easier to customize, because not having to build satis for every CSS change is a huge advantage for that matter.

So, coming back to your issue, I guess either your browser or your server had he HTML document cached.

staabm:
> @staabm Satis does not reference stylesheets, JS or icons, it embeds them in the HTML response. I was going to separate assets in this PR, but decided to go for that in a follow-up PR, focused on making satis easier to customize, because not having to build satis for every CSS change is a huge advantage for that matter.

Totally agree

> So, coming back to your issue, I guess either your browser or your server had he HTML document cached.

Maybe a workaround with the Clear-Site-Data header could work.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data
But maybe its also more like an edge case which shouldnt be considered.

Thx for the followup

PR #520, opened 2019-01-01
Updated Dockerfile (added libzip-dev)

Ilyes512:
Added build dependency libzip-dev which is needed when installing the zip php extension for PHP7.3.

Since satis is using the image php:7-cli-alpine (https://github.com/composer/satis/blob/master/Dockerfile#L18) which will now be php 7.3 on a new build.

Without adding this package you got the fallowing error while trying to build the image (partial output):

```bash
checking for zip archive read/writesupport... yes, shared
checking for the location of libz... no
checking pcre install prefix... no
checking libzip... yes
checking for the location of zlib... /usr
checking for pkg-config... /usr/bin/pkg-config
checking for libzip... not found
configure: error: Please reinstall the libzip distribution
```


alcohol:
Could you make the change similar to https://github.com/composer/docker/commit/5c3ca0eb892de3bf101d47004f83b1a0af40b758 ?

Ilyes512:
@alcohol Amended the commit and force pushed it! :)

PR #523, opened 2019-02-05
index: use configured url, rather the hardcoded

### LINKED ISSUE #522: hardcoded link to root
https://github.com/composer/satis/blob/6d7ba9292ce6d9ac26af9c21c147cdb063e4584d/views/index.html.twig#L22-L24

```twig
            <a href="/">
                <h1>{{ name|default('Composer repository') }}</h1>
            </a>
```

imho there's config key `homepage` which should be used here, instead of hardcoding.

glensc:
fixes #522

cc @alcohol

### DIFF
--- a/views/index.html.twig
+++ b/views/index.html.twig
@@ -19,7 +19,7 @@
 <div class="container">
     <header class="mt-3">
         <div class="d-flex justify-content-between align-items-center">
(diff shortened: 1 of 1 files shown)


alcohol:
Thanks :-)

PR #525, opened 2019-02-09
fix an invalid package name, preparing for Composer 2.0

linuxjuggler:
When installing the packages we get the following warning:

> Deprecation warning: require-dev.mikey179/vfsStream is invalid, it should not contain uppercase characters. Please use mikey179/vfsstream instead. Make sure you fix this as Composer 2.0 will error.

This small PR fix the issue

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "require-dev": {
         "phpunit/phpunit": "^6.4",
         "phpunit/phpunit-mock-objects": "^4.0",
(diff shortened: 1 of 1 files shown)


staabm:
Is composer still able to resolve this, even when the upstream package defines its name differently (with upper case chars?)

linuxjuggler:
with version 1.8 it is, but starting from Composer 2.0 it will not work

Issue #419, opened 2017-06-04
Is it possible to have Satis generate the GitLab source URLs as https instead of git@


Jacco-V:
When generating the package info `.json` file, Satis automatically generates URLs in the for of `git@gitlab.com:Foo/Bar.git`

```json
{"packages": {
   "foo/bar": {
      "dev-master": {
         "name": "Foo/Bar",
         "version": "dev-master",
         "version_normalized": "9999999-dev",
         "source": {
            "type": "git",
            "url": "git@gitlab.com:Foo/Bar.git",
            "reference": "a2c204bb385d552057bd07a4f19e6b91240c217c"
         },
```
**My question is:** Is there some sort of setting (such as `github-protocols`, but then for GitLab instead of GitHub) that would let Satis generate the https URL instead, so that it would look like:

```json
{"packages": {
   "foo/bar": {
      "dev-master": {
         "name": "Foo/Bar",
         "version": "dev-master",
         "version_normalized": "9999999-dev",
         "source": {
            "type": "git",
            "url": "https://gitlab.com/api/v3/projects/Foo%2FBar",
            "reference": "a2c204bb385d552057bd07a4f19e6b91240c217c"
         },
```

**Why this request?**  
Well, the `https` version can be downloaded with a 'personal access Token' for authentication, while the `git` version requires SSH keys, which are much harder to set up and distribute.

alcohol:
How did you define your source repositories?

Jacco-V:
The source repositories are defined in the `satis.json` file:

```json
{
    "name": "Project.com - Composer Packages",
    "homepage": "https://packages.Project.com",
    "output-dir": "/home/projectPackages/app/satisCustom/output",
    "repositories": [
        { "type": "vcs", "url": "https://gitlab.com/Project/Server" },
        { "type": "vcs", "url": "https://gitlab.com/Project/Core" },
        { "type": "vcs", "url": "https://gitlab.com/Project/Console" },
        { "type": "vcs", "url": "https://gitlab.com/Project/PHPClient" },
        { "type": "vcs", "url": "https://gitlab.com/Project/LaravelClient" }
    ],
    "require-all": true,
	"twig-template": "/home/projectPackages/app/satisCustom/views/index.html.twig",
    "config": {
        "gitlab-token": {
            "gitlab.com": "XXXXXXXXXXXXXXXXXXXX"
        }
    }
}
```

alcohol:
[...]

PR #527, opened 2019-02-25
Allow alternative VCS driver to be specified (e.g., for gitlab)

bradjones1:
Composer allows for alternative VCS drivers; satis could easily support this in the `add` command. This PR adds an optional option to this command, defaulting to the current `vcs` value, for users to take advantage of this feature.

https://getcomposer.org/doc/05-repositories.md#git-alternatives

Refs https://github.com/composer/satis/issues/419

### DIFF
--- a/src/Console/Command/AddCommand.php
+++ b/src/Console/Command/AddCommand.php
@@ -21,6 +21,7 @@ use Composer\Repository\VcsRepository;
 use Symfony\Component\Console\Helper\FormatterHelper;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


bradjones1:
@alcohol You are correct. Thanks.

bradjones1:
@alcohol In setting up our satis workflow, I did actually run into a case where this may be necessary; in our case, we do a `satis add` (actually called from a symfony controller) however the only access for Satis to GitLab is via the token.

When satis adds the repository, the type is the default, `vcs`, and the git driver attempts to do the clone, hanging us. I would need to specify `gitlab` at runtime to force this driver to fetch the info and validate the repository.

I am open to other workflows (perhaps the add command does as advertised and just adjusts the json, instead of doing any validation?) but for the moment this is necessary for us to get the repo into `satis.json`. Once it's there, the driver is properly selected.

bradjones1:
So this isn't necessary however I wonder if there is room for some documentation improvements either here or on composer core. Basically, in `::isRepositoryValid()`, a config is yielded from `Factory::createConfig()`, which does not parse any config options from `satis.json` but it will bring them in from `COMPOSER_HOME/config.json`. So any gitlab options must be in the config, even though on other operations satis will use the values in `satis.json`.

alcohol:
I cannot tell if `isRepositoryValid` is part of Composer or part of Satis, I'd have to check. If it is the former there isn't much we can do about this because Composer is not aware of Satis. If however if it is part of Satis, we could probably make adjustments there to allow for some sort of merge behaviour of configurations.

bradjones1:
@alcohol https://github.com/composer/satis/blob/7ab2986770b2106fc67d754bf4f3144065e8ff1b/src/Console/Command/AddCommand.php#L115

Thanks!

alcohol:
Is this change still desired?

bradjones1:
It is not a blocker for us as we've worked around it, but yes I think implementing a consistent behavior (re: the items raised in https://github.com/composer/satis/pull/527#issuecomment-469922768) would be good to do in Satis.

PR #528, opened 2019-02-26
Fix stale composer.lock content-hash

felixbuenemann:
This fixes the following warning when running `composer install`:

    Warning: The lock file is not up to date with the latest changes in composer.json.


felixbuenemann:
The stale lockfile was caused by 3c04234.

alcohol:
Thanks :-)

Issue #442, opened 2017-11-22
Satis doesn't support stability per package


avplab:
I am not sure this is an issue, but I couldn't find any official documentation where it's described. I would like to have `minimum-stability: stable` option configured for all packages, but also be able to provide an exception for specific package(I have to download a `-rc1` package). But `@RC` doesn't work in case of satis. After the source code [investigation](https://github.com/composer/satis/blob/master/src/PackageSelection/PackageSelection.php#L161)
 I found out that stability flags are not supported intentionally(so I suppose it's not an issue). Which alternative could be provided here?

stof:
@alcohol please close this issue, as it has been implemented

avplab:
Thank you

alcohol:
Sorry, I'm used to using the #closes annotation in commit messages myself :sweat_smile:

PR #533, opened 2019-03-19
Added "minimum-stability-per-package" config option

EfeloPHP:
This solve the issue #442, with a new config option named "minimum-stability-per-package" where we can override the "minimum-stability" option for a set of given packages : 

`"minimum-stability": "stable",
"minimum-stability-per-package": [
    "vendor/package-alpha": "dev"
],`

If it's ok, can I propose the feature for the 1.x branch also ?


alcohol:
@EfeloPHP sure

staabm:
should #442 be closed now?

stof:
yes, it should

PR #540, opened 2019-05-02
html: sort keywords

glensc:
sort `Keywords` list, because it shouldn't matter in what order they came on, output should be consistent.

- doc: https://twig.symfony.com/doc/2.x/filters/sort.html
- example: https://github.com/eventum/composer/commit/cd6d5731f437efb69133cfe6fba9d784eea6d6d4

background: I switched from original `vcs` or `pear` type to `composer` type, and the order changed: https://github.com/eventum/composer/commit/23c24e6338ab748f9277a007eef111519c93b678


alcohol:
Thanks

PR #542, opened 2019-05-03
exclude non-source from git export

glensc:
excludes configs for maintenance tools and tests.

could exclude docs and readme as well, but that is usually debatable.

I personally have an opinion in this matter:
- exclude all that is not needed runtime
- if you need tests, docs, use `--prefer-source` or git checkout the project yourself

to test outcome:
```
git archive --format=tar HEAD | tar tv
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,5 +15,6 @@
 /.git* export-ignore
 /Dockerfile export-ignore
 /box.json export-ignore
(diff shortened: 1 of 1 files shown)


alcohol:
I think adding the `docs` folder makes sense because it is there for github pages primarily.

alcohol:
Thanks

PR #543, opened 2019-05-03
version: back to 1.0.0-dev

glensc:
I was surprised when I updated my local satis code to dev-master (2845708) it said 2.0.0-dev

looks like that such major bump is very far ahead, so why just stick to 1.x for now.

the change was added in unrelated commit 7297e614900f6616498bc05444f852efb27bccaf and the pull request itself is also rather bogus, no defined scope: https://github.com/composer/satis/pull/538 not even clear was the change intended.

### DIFF
--- a/src/Satis.php
+++ b/src/Satis.php
@@ -15,5 +15,5 @@ namespace Composer\Satis;
 
 class Satis
 {
-    const VERSION = '2.0.0-dev';
+    const VERSION = '1.0.0-dev';
 }


alcohol:
Uhm, that change actually should have been done a long time ago when https://github.com/composer/satis/commit/d933c98209da62f8391a6c7098e759d83c2a02ca#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780 was committed (which was because several BC changes were introduced around that time).

PR #547, opened 2019-06-12
WIP: only-dependencies

hsegnitz:
This adds a config-flag for a usecase that cannot be done differently in an automatic way:

The parameter "only-dependencies" will instruct satis to NOT build the root level packages, only the ones for dependencies.
This is for the following use-case:
Using satis for speeding up continuous integration or other build pipelines behind a slow internet connection. In this case you want satis to automatically detect what packages are used by the software projects - but not have it build those packages as they will clutter up a lot of space and take a lot of time. But you also don't want to keep track of packages or versions manually. Like this you will only have to specify the project's repositories in satis.json and the project names and the new flag, the rest is done automatically.

Example:
```
{
  "name":     "example/projectmirror",
  "homepage": "https://satis.example.org",
  "repositories": [
    { "type": "composer", "url": "https://packagist.org" },
    { "type": "composer", "url": "https://satis.example.org/private-packages" },

    { "type": "vcs", "url": "git@git.example.com:my/project1.git" },
    { "type": "vcs", "url": "git@git.example.com:my/project2.git" }
  ],
  "require": {
    "my/project1":          "*",
    "my/project2": "*"
  },
  "output-dir": "/var/www/satis",
  "require-dependencies"    : true,
  "require-dev-dependencies": true,
  "only-dependencies":        true,
  "providers":                true,
  "archive": {
    "directory": "dist",
    "prefix-url": "https://satis.example.org/projectmirror",
    "skip-dev": false,
    "checksum": false
  }
}
```

WIP because I need to figure out how to write a unit test for this. - Help is of course gladly welcome.

### DIFF
--- a/src/PackageSelection/PackageSelection.php
+++ b/src/PackageSelection/PackageSelection.php
@@ -53,6 +53,9 @@ class PackageSelection
     /** @var bool required dev-dependencies if true. */
     private $requireDevDependencies;
 
+    /** @var bool do not build packages only dependencies */
+    private $onlyDependencies;
+
     /** @var bool Filter dependencies if true. */
     private $requireDependencyFilter;
 
@@ -101,6 +104,7 @@ class PackageSelection
         $this->requireAll = isset($config['require-all']) && true === $config['require-all'];
         $this->requireDependencies = isset($config['require-dependencies']) && true === $config['require-dependencies'];
         $this->requireDevDependencies = isset($config['require-dev-dependencies']) && true === $config['require-dev-dependencies'];
+        $this->onlyDependencies = isset($config['only-dependencies']) && true === $config['only-dependencies'];
         $this->requireDependencyFilter = (bool) ($config['require-dependency-filter'] ?? true);
 
         if (!$this->requireAll && !isset($config['require'])) {
@@ -166,7 +170,9 @@ class PackageSelection
             }
         }
 
-        $this->addRepositories($pool, $repos);
+        if (! $this->onlyDependencies) {
+            $this->addRepositories($pool, $repos);
+        }
 
         // determine the required packages
         $rootLinks = $this->requireAll ? $this->getAllLinks($repos, $this->minimumStability, $verbose) : $this->getFilteredLinks($composer);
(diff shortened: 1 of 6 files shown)


alcohol:
Can't you achieve this using the blacklist already? Or does Satis then not process the requirements of the blacklisted packages?

hsegnitz:
Hmmm... you are absolutely right, In my tests I wasn't using the blacklist the right way, my bad -.-

However, there would be a small difference:
* the PR removes the need to specify the package names twice
* the PR should create a package if it's also used as a dependency and not just the root-level entrypoint

But not having both would be okay for me - the redundancy isn't that bad with < 100 projects - and since I build the packages fully in a secondary satis index, I have them available...

So I leave it open for discussion...

alcohol:
Honestly, Satis is fully user maintained. The Composer team spends virtually no time on Satis anymore. We merely try to keep things up to date when necessary. If you feel this PR provides functionality Satis not yet fully offers, I am happy to merge it. Tests are a much appreciated bonus, but not a requirement.

hsegnitz:
Back home, I just verified my assumption, my "feature" loads the versions of a root lever repo which are in use as opposed to the blacklist ignoring them.

-> added unit test as well...

src/PackageSelection/PackageSelection.php
```diff
@@ -636,10 +640,12 @@ private function selectLinks(Pool $pool, $links, bool $isRoot, bool $verbose)
                 $uniqueName = $package->getUniqueName();
                 // add matching package if not yet selected
                 if (!isset($this->selected[$uniqueName])) {
-                    if ($verbose) {
-                        $this->output->writeln('Selected ' . $package->getPrettyName() . ' (' . $package->getPrettyVersion() . ')');
+                    if (false === $isRoot || $this->onlyDependencies === false) {
```
alcohol: Maybe inverse it combined with a continue statement? Reduces level of nesting.

SvenRtbg:
I'm not voting against anything, but I observe a constant inflow of suggestions and pull requests that all deal with an individual use case. They are coming about every half a year, so one could be happy that the community hasn't lost interest, on the other hand every new boolean config flag effectively doubles the amount of possible configuration combinations.

I don't think there are tests or even written expectations of how the different configurations should behave when enabled together. I don't want to spoil this particular PR or put on it the unexpected burden of writing all the missing tests, but I'd like to hear any statement about this from you.

alcohol:
Not sure what kind of statement you are looking for. Like I said, Satis is user maintained. We will accept most, if not all PRs, provided they are somewhat sensible and add functionality.

PR #549, opened 2019-07-06
Only match best candidates in dependencies, if desired.

simesy:
We want to restrict dependency packages to those that allow the parent packages to be built. This reduces the size of the output from 27MB to 700KB and allows us to provide a "white-listed packages" repository.

This is essentially an updated PR of https://github.com/composer/satis/pull/392 but I haven't encountered the issues the author of that PR encountered, so the code is fairly simple right now.


alcohol:
Cheers

PR #551, opened 2019-07-10
Update only one package

kpitn:
Actually you can't update only one package. (See this discussion #40)

With this PR you can select one package to update.

You need to define your json schema with a new "name" node :

    "repositories": [
        {
            "name": "package_name",
            "type": "vcs",
            "url":  "git@bitbucket.org:team/package_name.git"
        },
Now you can execute this query and it will only update one repository

bin/satis build my-schema.json my-webroot my_package_name

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -6,4 +6,327 @@ title: Using
 
 # Using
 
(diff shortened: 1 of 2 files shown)


kpitn:
@alcohol Sorry I delete the old PR, when try to rebase with master.

I update the documentation.

PR #555, opened 2019-07-24
Add name option to AddCommand

glickel:
Since the PR #551 I think it would be useful to add the name option to the AddCommand


alcohol:
Thanks

PR #558, opened 2019-08-01
Docs update for whitelist/blacklist wildcard match

SunMar:
Adds the availability of using a `*` for wildcard matching in the whitelist and blacklist to the documentation (wildcard matching was added in #434).


alcohol:
Thanks

PR #561, opened 2019-09-13
Blacklist to prevent undesirable packages from dependencies.

simesy:
So far "only-best-candidates" is working well for us. We provide this as the lone repository source for applications, and we have a high degree of confidence about the packages being used.

"only-best-candidates" effectively asks composer to choose the best candidate from a list, which is usually the latest within our high level constraints. However, sometimes a deep dependency causes a newer package to be selected (in addition to a desired package) that we don't want to list.

In such cases we might get both `package/bbb:1.1` (checked, security reviewed, and desired) and `package/bbb:1.2` (selected via a deep dependency, but not desired because it hasn't had a security review, and but which is anyway satisfied by `1.1`).

This new `blacklist` feature allows us to effectively define a blacklist policy in the top level of satis.json. To fix the situation described above we can:

```
    "require": {
        "package/aaa": "6.6.6"
    },
    "blacklist": {
        "package/bbb": ">1.1"
    },
    "require-dependencies": true,
    "only-best-candidates": true
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -135,6 +135,10 @@ If true, will only resolve and add dependencies, not the root projects listed in
 
 Useful with require-dependencies, returns a minimal set of dependencies resulting in a constrained package list.
 
+### blacklist
+
+Define a list of packages and versions to suppress in the final packages list. Takes the same format as the `require` section.
+
 ### require-dependency-filter
 
 If false, will include versions matching a dependency.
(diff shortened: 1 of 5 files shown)


alcohol:
If you could include a test, I will happily merge this asap.

simesy:
I will write a test thank you.

simesy:
Ready for review @alcohol - thanks for the attention.

alcohol:
Thanks for taking the time to write a test :-)

PR #583, opened 2020-02-03
Add PHP version requirement statement

### LINKED ISSUE #582: PHP version documentation outdated
https://getcomposer.org/doc/00-intro.md
The documentation states PHP version 5.3.2+ is required
`Composer requires PHP 5.3.2+ to run`

However the default installation is via composer and requires PHP 7.2+
https://github.com/composer/satis/blob/master/composer.json
```    
"require": {
        "php": "^7.2",
```

Actual behaviour on a stock CentOS 7 box:
```
$ composer create-project composer/satis:dev-master
[InvalidArgumentException]
Could not find package composer/satis with stability stable in a version installable using your PHP version 5.4.16.
```

If the documentation could be updated to mirror the composer.json file it should prevent others following me into building a CentOS 7 server defaulting to PHP 5.4 which can't then install composer.
If the docs were within the project I'd happily have issued a PR but don't believe that to be the case.

SvenRtbg:
Fixes #582.
Intentionally generic wording to prevent constant update sync issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ Simple static Composer repository generator.
 
 ## Run from source
 
+Satis requires a reasonable recent PHP version, it does not run with unsupported PHP versions.
+
 - Install satis: `composer create-project composer/satis:dev-master`
 - Build a repository: `php bin/satis build <configuration-file> <output-directory>`
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Simple static Composer repository generator.
 
 ## Run from source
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,8 @@ Simple static Composer repository generator.
 
 ## Run from source
 
+Satis requires a reasonable recent PHP version, it does not run with unsupported PHP versions.
```
alcohol:
```suggestion
Satis requires a recent PHP version, it does not run with unsupported PHP versions. Check the `composer.json` file for details.
```

SvenRtbg: Sure.

Issue #590, opened 2020-04-01
Required PHP version should be >=7.3


DavidGarciaCat:
Hello there,

When I create a new PHP 7.2's Docker Container and try to install Satis through `create-project composer/satis:dev-master` then I get an error because of the required PHP versions of the dependencies are 7.3 instead of 7.2

https://github.com/composer/satis/blob/4ffdaed2aaaf3daf1e4ce1b422d968f8cdc756f9/composer.json#L29

```
$ composer create-project composer/satis:dev-master
Creating a "composer/satis:dev-master" project at "./satis"
Installing composer/satis (dev-master 4ffdaed2aaaf3daf1e4ce1b422d968f8cdc756f9)
  - Installing composer/satis (dev-master 4ffdaed): Cloning 4ffdaed2aa from cache
    4ffdaed2aaaf3daf1e4ce1b422d968f8cdc756f9 is gone (history was rewritten?)
    Failed to download composer/satis from source: Failed to execute git checkout '4ffdaed2aaaf3daf1e4ce1b422d968f8cdc756f9' -- && git reset --hard '4ffdaed2aaaf3daf1e4ce1b422d968f8cdc756f9' --

fatal: failed to read object 4ffdaed2aaaf3daf1e4ce1b422d968f8cdc756f9: Operation not permitted

    Now trying to download from dist
  - Installing composer/satis (dev-master 4ffdaed): Downloading (100%)
Created project in /opt/project/satis
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for phpunit/php-code-coverage 8.0.0 -> satisfiable by phpunit/php-code-coverage[8.0.0].
    - phpunit/php-code-coverage 8.0.0 requires php ^7.3 -> your PHP version (7.2.29) does not satisfy that requirement.
  Problem 2
    - Installation request for phpunit/php-file-iterator 3.0.0 -> satisfiable by phpunit/php-file-iterator[3.0.0].
    - phpunit/php-file-iterator 3.0.0 requires php ^7.3 -> your PHP version (7.2.29) does not satisfy that requirement.
  Problem 3
    - Installation request for phpunit/php-invoker 3.0.0 -> satisfiable by phpunit/php-invoker[3.0.0].
    - phpunit/php-invoker 3.0.0 requires php ^7.3 -> your PHP version (7.2.29) does not satisfy that requirement.
  Problem 4
[...]

PR #591, opened 2020-04-02
Require PHP 7.3

DavidGarciaCat:
This should fix https://github.com/composer/satis/issues/590

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     },
     "bin": ["bin/satis"],
     "require": {
-        "php": "^7.2||^8.0",
+        "php": ">=7.3||^8.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
     },
     "bin": ["bin/satis"],
     "require": {
-        "php": "^7.2||^8.0",
+        "php": ">=7.3||^8.0",
```
SvenRtbg, line 29: Should probably be kept as `^7.3||^8.0`, otherwise mentioning 8.0 would be redundant.

DavidGarciaCat: Fair point! Pushing an update now

alcohol:
Thanks

Issue #594, opened 2020-04-14
Upgrade to composer 2.0 format


GromNaN:
**Is your feature request related to a problem? Please describe.**
Composer 2.0 is being actively developed, providing a new repository format.
https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-repository-implementors 

**Describe the solution you'd like**
The upgrade can be done in several steps:

- [x] #645 Upgrade dependency to `"composer/composer": "^2.0"` and `"composer/semver": "^2.0"`. ~(`dev-master` while development is not completed)~. And adapt code to all the breaking changes. 
-  [x] #596 Create a new `Builder` to generate the `/p2/%package%.json` files for each package, and add the `"metadata-url": "/p2/%package%.json"` to the main `packages.json`.
-  [ ] Add an option to minify with `Composer\Util\MetadataMinifier::minify()`.
-  [x] Add an option to dump a restrictive list of packages in `"available-packages": [...]`
-  [x] Add the [`providers-api`](https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#providers-api).

**Additional context**
Maybe @naderman, @Seldaek or @alcohol already planned to upgrade Satis on their own ?

stof:
Development on Satis is mostly done by the community. So contributions are welcome

alcohol:
Yeah I have to second what @stof said. We don't actively develop Satis ourselves. It depends entirely on community contributions.

designermonkey:
Satis without Composer v2 support no longer works for github authentication as the token format has changed. Please see here: https://github.com/composer/composer/issues/9907

I'm leaving this here so it's recorded somewhere, not to poke at the maintainers.

stof:
Composer 1 also supports the new github token format. But you need to use an uptodate composer 1.

designermonkey:
Noted. I don't know why composer on my satis server isn't updating to the latest v1 version.

alcohol:
> Noted. I don't know why composer on my satis server isn't updating to the latest v1 version.

What version (or commit ref) of satis are you running?

designermonkey:
It's an ansible thing, not the fault of this repo. Not to worry about it here :)

kvelaro:
Hi there, @alcohol 
We have annoying warning from packagist, and it seems in near feature we will face unpleasant problem.

[...]

PR #595, opened 2020-04-14
WIP: Upgrade to composer 2.0

GromNaN:
This is a work in progress.

There is a massive amount of changes in the core classes of Composer that needs to be applied to Satis code. https://github.com/composer/composer/compare/1.10...master

#594

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
     "bin": ["bin/satis"],
     "require": {
         "php": "^7.3||^8.0",
-        "composer/composer": "^1.9",
+        "composer/semver": "^2.0@dev",
+        "composer/composer": "^2.0@dev",
         "symfony/console": "^4.4||^5.0",
         "twig/twig": "^2.12||^3.0"
     },
@@ -36,7 +37,6 @@
     },
     "require-dev": {
         "ext-json": "*",
-        "friendsofphp/php-cs-fixer": "^2.15",
         "mikey179/vfsstream": "^1.6.7",
         "phpunit/phpunit": "^8.2||^9.0"
     },
--- a/tests/Builder/PackagesBuilderDumpTest.php
+++ b/tests/Builder/PackagesBuilderDumpTest.php
@@ -15,6 +15,7 @@ namespace Composer\Satis\Builder;
 
 use Composer\Json\JsonFile;
 use Composer\Package\Package;
+use Composer\Util\Loop;
 use org\bovigo\vfs\vfsStream;
 use org\bovigo\vfs\vfsStreamDirectory;
 use PHPUnit\Framework\TestCase;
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -36,7 +37,6 @@
     },
     "require-dev": {
         "ext-json": "*",
-        "friendsofphp/php-cs-fixer": "^2.15",
```
alcohol, line 39: Not sure I understand the reason for removing php-cs-fixer ? We use it for most of our linting and automated lint fixes.

alcohol: I guess it conflicts with the composer packages?

GromNaN: Yes, php-cs-fixer requires `"composer/semver": "^1.4"`.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.16/composer.json#L20

tests/Builder/PackagesBuilderDumpTest.php
```diff
@@ -15,6 +15,7 @@
 
 use Composer\Json\JsonFile;
 use Composer\Package\Package;
+use Composer\Util\Loop;
```
stof, line 18: why this change ?

stof:
For now, this upgrade the package to use Composer 2 internally. That's not the same than outputting the new metadata format, right ?

stof:
btw, for now, we should generate both the old and new format, so that the generated repository stays compatible with composer 1.

tests/Builder/PackagesBuilderDumpTest.php
GromNaN: An error.

GromNaN:
For now, I'm just fighting to update to `composer/composer: 2.0@dev`. But I'm getting lost by the new async mecanism of the `HttpDownloader`.

stof:
well, I think this PR should be considered separate from #594. We should not need to use composer 2 internally to output the new format (note that Packagist uses Composer 1 internally too)

tests/Builder/PackagesBuilderDumpTest.php
andrerom:
```suggestion
```

andrerom:
Now that 2.0 is stable, once #619 is in place it might be feasible to finish this one right?

composer.json
owenvoke: It looks like `php-cs-fixer` can be re-added as it supports `composer/semver 2.x` since `v2.15.9` 👍🏻

owenvoke:
Was this accidentally closed when the `master` branch was deleted (due to the base branch being `master`), or is development stopping on this? 🙂 If I can help at all with testing, let me know. 👍🏻

alcohol:
I think it was closed because the `master` branch was removed, yes.

ihor-sviziev:
@GromNaN, could you re-create a PR with the same changes to the `main` branch? Hope it will speed up the review and testing phase

GromNaN:
I'm working on it.

GromNaN:
New PR: https://github.com/composer/satis/pull/645

PR #596, opened 2020-04-19
Support composer 2.0 output format

GromNaN:
Adds a new option to output packages following the new [composer 2.0 format](https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-repository-implementors). Keep the existing package dumps for compatibility with composer 1.

Example for this simple `satis.json`:
```json
{
  "name": "corp/satis",
  "homepage": "http://127.0.0.1",
  "repositories": [
    {"type": "vcs", "url": "https://github.com/php-fig/log"},
    {"type": "vcs", "url": "git@github.com:Seldaek/monolog.git"}
  ]
}
```

Output structure:
image

Notes:
- The new option `"minified": "composer/2.0"` has not been implemented, since it requires `composer 2.0`.
- `"available-packages"` can be implemented later.
- [`provider-api`](https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#providers-api) can be implemented later.

### DIFF
--- a/src/Builder/PackagesBuilder.php
+++ b/src/Builder/PackagesBuilder.php
@@ -16,6 +16,7 @@ namespace Composer\Satis\Builder;
 use Composer\Json\JsonFile;
 use Composer\Package\Dumper\ArrayDumper;
 use Composer\Package\PackageInterface;
+use Composer\Semver\VersionParser;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class PackagesBuilder extends Builder
@@ -75,6 +76,42 @@ class PackagesBuilder extends Builder
             }
         } else {
             $repo['includes'] = $this->dumpPackageIncludeJson($packagesByName, $this->includeFileName);
+        }
+
+        if (isset($this->config['composer-2.0']) && $this->config['composer-2.0']) {
+            $metadataUrl = 'p2/%package%.json';
+            if (!empty($this->config['homepage'])) {
+                $repo['metadata-url'] = parse_url(rtrim($this->config['homepage'], '/'), PHP_URL_PATH) . '/' . $metadataUrl;
+            } else {
+                $repo['metadata-url'] = $metadataUrl;
+            }
+
+            // Dump the packages
+            foreach ($packagesByName as $packageName => $versionPackages) {
+                $stableVersions = [];
+                $devVersions = [];
+                foreach ($versionPackages as $version => $versionConfig) {
+                    if ('stable' === VersionParser::parseStability($versionConfig['version'])) {
+                        $stableVersions[] = $versionConfig;
+                    } else {
+                        $devVersions[] = $versionConfig;
+                    }
+                }
+
+                // Stable versions
+                $dumpPackages = [$packageName => $stableVersions];
+                $this->dumpPackageIncludeJson(
+                    $dumpPackages,
+                    str_replace('%package%', $packageName, $metadataUrl)
+                );
+
+                // Dev versions
+                $dumpPackages = [$packageName => $devVersions];
+                $this->dumpPackageIncludeJson(
+                    $dumpPackages,
+                    str_replace('%package%', $packageName.'~dev', $metadataUrl)
+                );
+            }
         }
 
         $this->dumpPackagesJson($repo);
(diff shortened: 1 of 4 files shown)


GromNaN:
Thank you @stof for [your comment](https://github.com/composer/satis/pull/595#issuecomment-615267505). Indeed we don't need to upgrade to `composer/composer: 2.0` to implement its new format.

alcohol:
Is this ready to be merged?

stof:
Should this even be a configurable option ? Why not always outputting the new format ?

GromNaN:
Thank you @stof for raising the question.

It's more a matter of backward compatibility than an actual option. I see many issues that might pop :
- if the /p2 directory is not accessible, composer would receive a 404 for every packages and consider the packages as not availables on the repository. 
- if the cache headers are not well configured to return a 304, package files will be downloaded every time.

GromNaN:
I've removed the option. It's ready.

alcohol:
> Should this even be a configurable option ? Why not always outputting the new format ?

Uhm.. because we don't really version Satis and maybe users are not yet ready to move to the new format? Or is it backwards compatible? I honestly have not read much yet into the 2.0 changes.

stof:
@alcohol this PR outputs the new format in addition to the old one, not as a replacement (we cannot drop the old format yet anyway, as Composer 1.x does not support the new format).

Note that this is the same approach used by Packagist: dump both formats. Composer 2 will use the new format if available (detected based on the new field in the main `package.json`)

stof:
>     * The new option `"minified": "composer/2.0"` has not been implemented, since it requires `composer 2.0`.

AFAIK, Composer 2.0 does not contain the code to write to the new format, only to read from it.
The place where this code exists right now is in Packagist.

@Seldaek any plan to release this minifier as a separate package, so that Packagist and Satis (and probably also Private Packagist) can share the implementation ?

src/Builder/PackagesBuilder.php
```diff
@@ -77,6 +79,38 @@ public function dump(array $packages): void
...
+            $repo['metadata-url'] = parse_url(rtrim($this->config['homepage'], '/'), PHP_URL_PATH) . '/' . $metadataUrl;
+        } else {
+            $repo['metadata-url'] = $metadataUrl;
+        }
+
+        foreach ($packagesByName as $packageName => $versionPackages) {
+            $stableVersions = [];
+            $devVersions = [];
+            foreach ($versionPackages as $version => $versionConfig) {
+                if ('stable' === VersionParser::parseStability($versionConfig['version'])) {
```
Seldaek, line 94: This is not correct, dev versions are only those with 'dev' stability, alpha/beta/etc goes in the other files with all the tags (non-dev versions, not really "stable" versions).

Seldaek:
It is published at https://github.com/composer/composer/blob/master/src/Composer/Util/MetadataMinifier.php#L43 - but as @GromNaN said, this requires composer v2 as dependency.

stof:
@Seldaek couldn't it be released as a separate package instead, to allow using it in code still using composer 1 (Satis and Packagist) ?

src/Builder/PackagesBuilder.php
stof: The idea of the split is that `dev` versions are invalidated at each push to the repo (well, at least one of them). Tags are immutable, and the non-dev file will then be updated only each time you do a release, which is much less often than each push.

Seldaek:
I'd rather not have another package just for this, you can copy paste the 20 lines of code until 2.0 stable if it's urgent, it's very unlikely to change anyway

GromNaN:
That's an optimization that is not required right now. We can wait for the upgrade to composer 2.0 (#595) to implement the minification.

src/Builder/PackagesBuilder.php
GromNaN: Thanks for the review and the explanation. The code has been updated.

alcohol:
@GromNaN ready to merge now?

GromNaN:
Yes, ready.

alcohol:
Cheers :tada:

glensc:
there is point of having p2 off:
- in private satis installation, you save some disk space by not providing p2 index

GromNaN:
Thank you for your feedback @glensc. There is 2 metadata files for each package. This maximum a few Mb. The archives take way more space.
```
162K p2/monolog/monolog.json
8.8K p2/monolog/monolog~dev.json
```

Once composer 2.0 will be released (soon), we should add an option to remove the old metadata format.

Issue #194, opened 2015-01-08
Suggestion: Add a favicon


valioDOTch:
Feature suggestion:
Integrate a favicon, so that's it's easier to find the link in the bookmarks

Suggested image template: 
http://commons.wikimedia.org/wiki/File:Package-x-generic.svg

Ico-File created from suggested template above with 48x48, 32x32 and 16x16px size http://s000.tinyupload.com/index.php?file_id=15029993054614563384
(sorry for the ads, but github doesn't support .ico-filetype for image uploads)

GromNaN:
This is a nice icon 
![Favicon](http://upload.wikimedia.org/wikipedia/commons/thumb/8/8f/Package-x-generic.svg/48px-Package-x-generic.svg.png)

 It could replace be used on Packagist.com too ![Favicon](https://packagist.org/favicon.ico)

valioDOTch:
Yes I think the image language is more obvious with this one

alcohol:
@seldaek I'll leave this up to you, as to what is an appropriate or acceptable image.

timdiels1:
So any progress on this?

alcohol:
I don't see any open pull requests, so I guess not.

timdiels1:
@alcohol I would love to help for this, but no clue where to start as i don't seem to find a way to test this locally...

alcohol:
The quickest way to test this would be to run Satis within docker with a very minimal configuration.

moneef:
@alcohol @valioDOTch @timdiels1 , i have created a pull request https://github.com/composer/satis/pull/597 to add favicon.

theimbender:
I'm pretty sure this issue can be closed, a favicon was added to satis more than two years ago.

PR #597, opened 2020-05-05
Adding a favicon to satis webpage

moneef:
Refs #194 
Icon source: http://commons.wikimedia.org/wiki/File:Package-x-generic.svg


views/index.html.twig
```diff
@@ -6,6 +6,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="robots" content="noindex,nofollow" />
+    <link rel="shortcut icon" type="image/png" href="/satis/views/assets/img/favicon.png">
```
GromNaN, line 9: This absolute path is specific to your deployment. The file must be duplicated into the "output" directory and the path must be relative.

alcohol: Yeah @GromNaN is correct. Satis is basically a static site generator. The output directory is configurable, but any assets should be placed there along with the generated html. Alternatively, you could inline the png like we did with most other javascript/css/image assets. See https://stackoverflow.com/a/5568484 for an example.

glensc:
For future generations add origin of the icon source (describe in PR body). So it could be updated or possible license issues can be addressed.

alcohol:
The source is referenced in the issue. True, it's indirect, but it is known.

glensc:
There is no issue reference, the PR body is completely empty.

alcohol:
As pedantic as ever. Anyway, I see the link clear as day on my screen. But there you go, edited the PR body so you can manage to find it too.

Issue #574, opened 2019-11-22
PackageSelection fails when a package json is an empty array.


solody:
```bash
composer_1  | Downloading https://packages.drupal.org/8/drupal/rules_ui%244f53cda18c2baa0c0354bb5f9a3ecbe5ed12ab4d8e11ba873c2f11161202b945.json
composer_1  | Writing /composer/cache/repo/https---packages.drupal.org-8/provider-drupal$rules-ui.json into cache
composer_1  | 
composer_1  | In ComposerRepository.php line 376:
composer_1  |                              
composer_1  |   [ErrorException]           
composer_1  |   Undefined index: packages  
composer_1  |                              
composer_1  | 
composer_1  | Exception trace:
composer_1  |   at /satis/vendor/composer/composer/src/Composer/Repository/ComposerRepository.php:376
composer_1  |  Composer\Util\ErrorHandler::handle() at /satis/vendor/composer/composer/src/Composer/Repository/ComposerRepository.php:376
composer_1  |  Composer\Repository\ComposerRepository->whatProvides() at /satis/vendor/composer/composer/src/Composer/DependencyResolver/Pool.php:204
composer_1  |  Composer\DependencyResolver\Pool->computeWhatProvides() at /satis/vendor/composer/composer/src/Composer/DependencyResolver/Pool.php:193
composer_1  |  Composer\DependencyResolver\Pool->whatProvides() at /satis/src/PackageSelection/PackageSelection.php:685
composer_1  |  Composer\Satis\PackageSelection\PackageSelection->selectLinks() at /satis/src/PackageSelection/PackageSelection.php:188
composer_1  |  Composer\Satis\PackageSelection\PackageSelection->select() at /satis/src/Console/Command/BuildCommand.php:184
composer_1  |  Composer\Satis\Console\Command\BuildCommand->execute() at /satis/vendor/symfony/console/Command/Command.php:255
composer_1  |  Symfony\Component\Console\Command\Command->run() at /satis/vendor/symfony/console/Application.php:934
composer_1  |  Symfony\Component\Console\Application->doRunCommand() at /satis/vendor/symfony/console/Application.php:273
composer_1  |  Symfony\Component\Console\Application->doRun() at /satis/src/Console/Application.php:49
composer_1  |  Composer\Satis\Console\Application->doRun() at /satis/vendor/symfony/console/Application.php:149
composer_1  |  Symfony\Component\Console\Application->run() at /satis/bin/satis:26
composer_1  | 
[...]

PR #603, opened 2020-06-04
fix-package-selection-broken

solody:
Issues#574 [https://github.com/composer/satis/issues/574](https://github.com/composer/satis/issues/574)

Fix package selection broken when making mirror for https://packages.drupal.org/8

### DIFF
--- a/bin/satis
+++ b/bin/satis
@@ -1,6 +1,10 @@
 #!/usr/bin/env php
 <?php
-
+ini_set ("memory_limit","-1");
+error_reporting(E_ERROR | E_WARNING | E_PARSE);
+//报告所有错误
+error_reporting(E_ALL);
+ini_set("display_errors","On");
 /*
  * This file is part of composer/satis.
  *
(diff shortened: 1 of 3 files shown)


bin/satis
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env php
 <?php
-
+ini_set ("memory_limit","-1");
+error_reporting(E_ERROR | E_WARNING | E_PARSE);
+//报告所有错误
+error_reporting(E_ALL);
+ini_set("display_errors","On");
```
SvenRtbg, line 7: Does this belong into your pull request? Non-englisch comment, increasing memory limit, double setting the error reporting, switching on error display all looks pretty much like leftovers from debugging.

src/PackageSelection/PackageSelection.php
```diff
@@ -30,6 +30,7 @@
 use Composer\Semver\Constraint\EmptyConstraint;
 use Composer\Semver\VersionParser;
 use Composer\Util\Filesystem;
+use PHPUnit\Exception;
```
SvenRtbg, line 33: I think you imported the wrong exception here.

SvenRtbg:
Additionally it would be nice if you'd be able to create any kind of test case for the fix.

alcohol:
Not sure what you are trying to achieve here, but this PR in its current state makes no sense to me. Going to close it.

PR #608, opened 2020-06-10
Docker: Upgrade based PHP container.

zeriyoshi:
PHP 7 has been EOL.
Bump based container PHP 7.0 to 7.4.

It works fine in my environment, included tests.


zeriyoshi:
Sorry, Wrong version rules. It closed.

PR #610, opened 2020-09-02
Avoid rewriting the same contents in a package file

GromNaN:
Since caching is done via the use of `If-Modified-Since` header, files should not be touched when not modified.
https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-repository-implementors


GromNaN:
I found that we should use [`Composer\Json\JsonFile`](https://github.com/composer/composer/blob/1.10/src/Composer/Json/JsonFile.php#L148) to handle this case.

PR #614, opened 2020-10-20
Set memory limit same as composer does

### LINKED ISSUE #532: Out of memory using PHP defaults
By default, at least for my installation of PHP, the `memory_limit` is set to `128M`. @alcohol previously mentioned in #437 that "Composer (the core component of Satis) itself requires roughly 1gb to be available at least." This is almost true. 

Composer actually manages the `memory_limit` itself to ensure it has enough memory to run ([`1.5G` to be more exact](https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors)).

I am requesting, as a feature/enhancement to satis, that like composer it should internally manage its memory requirements (as well as detecting OOM errors and pointing people to the [aforementioned composer documentation](https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors)).

Thank you.

freezy-sk:
Set memory limit same as Composer does.
Respects env var SATIS_MEMORY_LIMIT or COMPOSER_MEMORY_LIMIT if set.
Fixes #532

### DIFF
--- a/bin/satis
+++ b/bin/satis
@@ -22,5 +22,39 @@ if ((!$loader = includeIfExists(__DIR__.'/../vendor/autoload.php')) && (!$loader
     exit(1);
 }
 
+if (function_exists('ini_set')) {
+    // Set user defined memory limit or use Composer's user defined memory limit
+    if ($memoryLimit = getenv('SATIS_MEMORY_LIMIT')) {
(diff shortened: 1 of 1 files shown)


alcohol:
Thanks

PR #619, opened 2020-10-29
Setup CI workflow

alcohol:


andrerom:
Nice, unsure if it is wanted, but could also could add validation of composer.json, optionally also adding CS fixers:
```
jobs:
  validate_composer_json:
    name: Composer & Satis config validation
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Validate composer.json & lock
        run: composer validate --strict

  php-cs-fixer:
    name: PHP-CS-Fixer
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: PHP-CS-Fixer
        uses: docker://oskarstark/php-cs-fixer-ga:2.16.6
        with:
          args: --config=.php_cs --dry-run --diff --show-progress=estimating

  js-cs-fixer:
    name: JS-CS-Fixer
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Install modules
        run: yarn install
      - name: Run ESLint
        run: yarn check-cs
```

zaherg:
Maybe also using `https://github.com/shivammathur/setup-php` to setup PHP instead of only depending on the preinstalled PHP versions?

andrerom:
👍 
If so, example from another project:
```
jobs:
  phpunit:
    name: Unit Tests
    runs-on: ${{ matrix.operating-system }}
    strategy:
      fail-fast: false
      matrix:
        operating-system: [ubuntu-latest]
        php-versions: ['7.3']
    steps:
      - name: Checkout
        uses: actions/checkout@v2

      # https://github.com/shivammathur/setup-php#cache-composer-dependencies
      - name: Get composer cache directory
        id: composercache
        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
      - name: Cache dependencies
        uses: actions/cache@v2
        with:
          path: ${{ steps.composercache.outputs.dir }}
          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
          restore-keys: ${{ runner.os }}-composer-

      - name: Setup PHP
        uses: shivammathur/setup-php@v2
        with:
          php-version: ${{ matrix.php-versions }}
          coverage: none
          tools: composer:v2
          extensions: xsl, intl, gd

      - name: Install Composer dependencies
        run: composer install --no-progress --prefer-dist --optimize-autoloader

      - name: Run Unit Tests
        run: php bin/phpunit
```

alcohol:
I prefer running PHP from a docker container over installing it. Right now I need to figure out why these specific tests keep failing though (they fail for me locally as well). Missing extension perhaps (but if that is the case, we should list it as a requirement in our composer.json).

PR #622, opened 2020-11-16
Allow relative paths

andrerom:
Why?
To be able to more easily deploy satis across localhost, staging/dev domain and production.

How?
Somehow allowing relative paths as it simplifies config handling a lot:
1. Make `homepage` optional _(as is the case in PR here currently)_ for anything but WebBuilder
    _TODO: If this approach is taken, then `InitCommand` should be relaxed to allow this_
2. Change homepage type to `uri-reference` which allows relative paths + handle duplicate slashes in code when set to `"\"`
    _PRO: This might be cleaner as it then treat the property the same across all code, and it can still be required._
3. ...?

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -3,7 +3,7 @@
     "name": "Repository Configuration",
     "type": "object",
     "additionalProperties": false,
-    "required": [ "name", "homepage" ],
+    "required": [ "name" ],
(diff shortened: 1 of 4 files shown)


andrerom:
Composer does not seem to like this very much once it comes to downloading archive, so dropping this for now.

PR #625, opened 2020-12-04
[Purge] Add --dry-run option

andrerom:
Very useful in order to see what files would be purged, to get an idea if satis.json requirements has changed over time and for which packages.


alcohol:
Thanks!

PR #626, opened 2020-12-04
[Build] Add support for SATIS_HOMEPAGE environment variable

andrerom:
If set, will be used over homepage settting in satis.json.
Usefull for staging servers or similar where it's difficult to know url during builds steps for instance before project is deployed.

### DIFF
--- a/src/Console/Command/PurgeCommand.php
+++ b/src/Console/Command/PurgeCommand.php
@@ -39,8 +39,9 @@ on given json file (satis.json is used by default) and the
 newest json file in the include directory of the given output-dir.
 
 In your satis.json (or other name you give), you must define
-"archive" argument.
+"archive" argument. You also need to define "homepage"* if you don't use archive "prefix-url".
 
+* "homepage" may be overloaded using SATIS_HOMEPAGE environment variable.
 EOT
             );
     }
@@ -75,7 +76,7 @@ EOT
 
         $prefix = sprintf(
             '%s/%s/',
-            $config['archive']['prefix-url'] ?? $config['homepage'],
+            $config['archive']['prefix-url'] ?? getenv('SATIS_HOMEPAGE') ?? $config['homepage'],
             $config['archive']['directory']
         );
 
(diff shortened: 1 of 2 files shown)


src/Console/Command/PurgeCommand.php
```diff
@@ -39,8 +39,9 @@ protected function configure()
 newest json file in the include directory of the given output-dir.
 
 In your satis.json (or other name you give), you must define
-"archive" argument.
+"archive" argument. You also need to define "homepage"* if you don't use archive "prefix-url".
```
yannoff, line 42: Is the extra leading `*` after `"homepage"` intentional ? :thinking:

andrerom: yes, it points to 2 lines below: ´* "homepage" may be overloaded using SATIS_HOMEPAGE environment variable.´

could probably be rewritten to `You also need to define "homepage" argument or "SATIS_HOMEPAGE" environment variable, if you don't use archive "prefix-url".`

and then drop the line below.

yannoff: Sounds clearer indeed :wink: :rocket:

src/Console/Command/PurgeCommand.php
```diff
@@ -39,8 +39,9 @@ protected function configure()
...
 
+* "homepage" may be overloaded using SATIS_HOMEPAGE environment variable.
```
andrerom, line 44: I don't have access to edit the original repo anymore, but that would be the following change:

```suggestion
"archive" argument. You also need to define "homepage" argument or "SATIS_HOMEPAGE" environment variable if you don't use archive "prefix-url" argument.
```

alcohol: Ah wait, I can't commit those suggested changes either cause they would be applied to the remote branch in a repository that I have no access to.

konradoboza (approved):

andrerom:
@alcohol Got some help from @konradoboza, so PR updated

PR #632, opened 2021-02-06
Add inline, base64-encoded favicon

theimbender:
Implements suggestion #194. My understanding is that pull request #597 to add favicon support was well-received but the issue was that the image asset used an absolute path. This change adds [`twigphp/html-extra`](https://github.com/twigphp/html-ext) to support the [`data_uri`](https://twig.symfony.com/doc/3.x/filters/data_uri.html) filter, so that a data URI can be generated and included within the Twig template.

Icon source: https://commons.wikimedia.org/wiki/File:Package-x-generic.svg


theimbender:
Got tests passing, I forgot to add the actual image file.

alcohol:
Rebased your branch and merged it into master.

views/index.html.twig
```diff
@@ -6,6 +6,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="robots" content="noindex,nofollow" />
+    <link rel="shortcut icon" type="image/png" href="{{ source('assets/img/favicon.png')|data_uri(mime="image/png") }}">
```
stof, line 9: should be `rel="icon"`: https://mathiasbynens.be/notes/rel-shortcut-icon

alcohol: > Update: For historical reasons, the HTML specification now allows the use of shortcut as a link relation if it’s immediately followed by a single U+0020 space character and the icon keyword. Of course, it’s still better not to use any HTML at all.

But it is probably better to keep it simple yes.

theimbender: Updated #634 to include change for this

PR #634, opened 2021-02-08
Front end performance improvements

theimbender:
This pull request does two things:

1. Update NPM dependencies to work with OSX Catalina and above (current version of `node-gyp` is incompatible), including updating config for webpack 5 to continue transpiling for IE11
2. Replace jQuery, including bootstrap collapse script, with vanilla js that is supported in IE11+. This reduces the overall page size by ~100KB, and improves lighthouse performance metrics by removing unused javascript. Testing on my local also showed 0.4s less time to first contentful paint.


package.json
```diff
@@ -1,12 +1,11 @@
...
   "devDependencies": {
-    "@symfony/webpack-encore": "^0.28.0",
-    "bootstrap": "^4.1.3",
-    "date-fns": "^2.0.0",
-    "jquery": "^3.3.1",
-    "node-sass": "^4.9.4",
+    "@symfony/webpack-encore": "^1.0.5",
+    "bootstrap": "^4.6.0",
+    "date-fns": "^2.17.0",
+    "node-sass": "^5.0.0",
```
stof, line 6: with node-sass being deprecated, I think migrating to the `sass` package would be better actually

views/assets/js/App/Collapse.js
```diff
@@ -0,0 +1,68 @@
...
+        if (instance.handleClick(event)) {
+          event.preventDefault();
+        }
+      });
+    });
+
+    // Keep track of the natural heights of the collapsed elements
+    this.collapsibleElements.forEach(function (element) {
+      const initiallyVisible =
+        Array.prototype.indexOf.call(element.classList, "show") !== -1;
```
stof, line 52: use `element.classList.contains("show")` instead. It will be a lot more readable.

stof:
I don't think the current code is compatible with IE 11. AFAICT, it does not enable babel to transpile classes to ES5.

alcohol:
With my knowledge of frontend development being quite limited, I'm going to defer to @stof here. Please see if his comments are something you can work on. If you need more feedback, don't hesitate to ask.

theimbender:
Was testing against IE11 in parallels and the web view was working both before and after my changes, classes do get transpiled. I did hit an issue where arrow functions weren't, but then when I set those webpack 5 environment settings it started working again. Wasn't sure what the supported browsers were for this project, but didn't want to be the one to decide not to support something that was already working.

Also removed `shortcut` from the favicon, was hoping to keep it for similar reasons but it wasn't displaying either way in IE11 and was net new functionality there so figured it wasn't worth continuing on that.

stof:
ah, I forgot that webpack-encore is always enabling babel (allowing to configure it, but not to disable it)

alcohol:
Can this be merged now?

theimbender:
All feedback has been implemented

alcohol:
Thanks

PR #636, opened 2021-02-11
Ability to toggle visibility of fields

### LINKED ISSUE #513: UI: toggle visibility of fields per entry
It would be nice if the following fields could be enabled/disabled on the fly, e.g. show or hide the "author" attribute of all packages.

- [ ] homepage
- [ ] license
- [ ] author
- [ ] support
- [ ] releases (?)

Not sure how this would be best represented visually, but I imagine the actual implementation could be 100% in the frontend/javascript.

theimbender:
Resolves #513 by adding a UI for toggling package fields. Button next to the package filter toggles the collapse/expand state of the checkbox form. Visibility settings are saved to localStorage so that they persist across sessions. Defaults to show all fields.

Screen Shot 2021-02-11 at 4 38 48 PM


alcohol:
Thanks!

PR #637, opened 2021-02-14
Add 'Filter by' field and update settings panel

theimbender:
## Notes
- Adds new "Filter by" field so you can pick which field to search within
- Change settings toggle from text to a gear icon ([source](https://commons.wikimedia.org/wiki/File:Octicons-gear.svg)) from same Octicons library as existing link icon, under MIT license
- Change search field `input` to be of `type=search` so the search term can be cleared natively by browsers
- Add `satis` prefix to localStorage item keys to prevent potential collisions on the same domain
- Fix IE11 regression where escape key wasn't clearing filter

## Screenshot
Screen Shot 2021-02-14 at 10 40 12 AM


alcohol:
Thanks. Perhaps as a separate PR, you could introduce an eslint / prettier configuration? That way we can ensure that future PRs (from other individuals) also maintain the same code and style guides.

PR #639, opened 2021-02-16
Resolves #317 - Provide code snippets per release

theimbender:
## Notes
- Implements functionality so that when you click on an individual release, a snippet of code is created for your project's `composer.json` for that specific release (issue #317)
- Provides copy to clipboard button, using [icon from Octicons](https://commons.wikimedia.org/wiki/File:Octicons-clippy.svg)
- Change badge class of keywords so they are more visually distinct from the new release badges

## Screenshot
Screen Shot 2021-02-16 at 6 32 10 PM

PR #640, opened 2021-02-17
Add badge wrapper for metapackage

theimbender:
Fixes missing formatting for metapackages [per comment](https://github.com/composer/satis/pull/639#discussion_r577503664) raised in #639, and also makes all release badges keyboard navigable


views/package.html.twig
```diff
@@ -94,13 +94,13 @@
...
                 {%- elseif version.distType -%}
                 <a class="badge badge-pill badge-secondary" href="{{ version.distUrl }}" title="dist-reference: {{ version.distReference }}{{ branch_alias }}">{{ version.prettyVersion }}</a>
                 {%- elseif version.sourceUrl matches '#^https?:\/\/#' -%}
-                <a class="badge badge-pill badge-secondary"  href="{{ version.sourceUrl }}" title="source-reference: {{ version.sourceReference }}{{ branch_alias }}">{{ version.prettyVersion }}</a>
+                <a class="badge badge-pill badge-secondary" href="{{ version.sourceUrl }}" title="source-reference: {{ version.sourceReference }}{{ branch_alias }}">{{ version.prettyVersion }}</a>
                 {%- else -%}
-                <span class="badge badge-pill badge-secondary"  title="source-reference: {{ version.sourceReference }}{{ branch_alias }}">{{ version.prettyVersion }}</span>
+                <span class="badge badge-pill badge-secondary" tabindex="0" title="source-reference: {{ version.sourceReference }}{{ branch_alias }}">{{ version.prettyVersion }}</span>
```
stof, line 103: why adding `tabindex=0` ? Why making it focusable when it is not a link ?

views/assets/css/style.scss
```diff
@@ -55,6 +55,10 @@ pre.file {
   }
 }
 
+// Copy code snippets
+.field-releases .badge {
+  cursor: pointer;
```
stof, line 60: Why making it a pointer when it is not a link ?

theimbender:
For both of those comments, all badges now have a click event that shows the code snippet text field and prepopulates it with the appropriate snippet. So they essentially _are_ links, or at least actionable, and should be treated as such to the user.

stof:
@theimbender how does this click event listener work when the version was already an actual link by itself. Wouldn't this conflict together ?

Also, from an accessibility point of view, a JS action should be triggered by a `<button type=button>`, not by a focusable `<span>`. Otherwise, screen readers won't know about them and they won't be usable from the keyword either.

theimbender:
Currently it prevents the link from navigating and a user would need to open the context menu to choose to open it in a new tab, or some other behavior. So that's a good point, it _would_ be unexpected.

Do we want to preserve the old behavior of just linking out, maybe have a setting where someone could choose between whether to expose the code snippets on click, or for releases to just be links?

For every repo I added to my satis test repository, all the links were identical across releases, they were just going to the same base URL of the repo, so I thought this was already broken.

theimbender:
Do you know of a package where the dist or source urls are different between releases so I could test this? Or what config I would need to change on a repo's `composer.json` to set up a test package?

theimbender:
Been reading the docs on source/dist in composer, i'll set up some more tests on my local this weekend and come up with a toggle because it seems like both workflows are valid

alcohol:
> Do you know of a package where the dist or source urls are different between releases so I could test this? Or what config I would need to change on a repo's `composer.json` to set up a test package?

Not a clue honestly. I almost never use source installs. I reckon they are always different though cause source refers to repository with commit hash usually, and dist often refers to github zip/tarball?


> Been reading the docs on source/dist in composer, i'll set up some more tests on my local this weekend and come up with a toggle because it seems like both workflows are valid

Appreciate the time you are putting in. But if things get too complex, I would perhaps argue that sometimes less is more.

theimbender:
Ok, that makes sense, I personally don't use source and I feel like being able to copy dependency snippets is more useful than linking out. I'll make those updates hopefully within the next few days, gotten swamped lately. Let me close this pr and I'll create a new one that specifically updates the front end to use buttons and more semantic markup.

PR #645, opened 2021-04-05
Upgrade to composer 2.0

GromNaN:


ihor-sviziev:
Hi @GromNaN,
Could you review and try to fix the test failures?

GromNaN:
This is a work in progress. Composer codebase as significantly changed and I try to understand the mechanism of both Composer and satis config.

src/PackageSelection/PackageSelection.php
```diff
@@ -180,34 +179,37 @@ public function select(Composer $composer, bool $verbose): array
...
 
         // determine the required packages
         $rootLinks = $this->requireAll ? $this->getAllLinks($repos, $this->minimumStability, $verbose) : $this->getFilteredLinks($composer);
 
         // select the required packages and determine dependencies
-        $depsLinks = $this->selectLinks($pool, $rootLinks, true, $verbose);
+        $depsLinks = $this->selectLinks($repositorySet, $rootLinks, true, $verbose);
 
         if ($this->requireDependencies || $this->requireDevDependencies) {
+            $repositorySet = new RepositorySet($this->minimumStability, $stabilityFlags);
```
stof, line 192: do we need to create a new one here ?

GromNaN: Yes, when a `Pool` has been created by the `RepositorySet` ([in `selectLinks`](https://github.com/composer/satis/pull/645/files#diff-d16a5020dc605f5cb12e0cf898bb6a7d2b2a7eacd2dc2bbc77b56b1d61d005a3R721)), it is locked. Then other repositories [cannot be added](https://github.com/composer/composer/blob/2.0.12/src/Composer/Repository/RepositorySet.php#L128).

src/PackageSelection/PackageSelection.php
```diff
@@ -711,32 +707,17 @@ private function selectLinks(Pool $pool, array $links, bool $isRoot, bool $verbo
...
-                    $matches = $pool->whatProvides($name, $link->getConstraint(), true);
+                    $matches = $repositorySet->createPoolForPackage($link->getTarget())->whatProvides($name, $link->getConstraint());
                 }
 
                 if (0 === \count($matches)) {
                     $this->output->writeln('<error>The ' . $name . ' ' . $link->getPrettyConstraint() . ' requirement did not match any package</error>');
                 }
             }
 
-            if (!$isRoot && $this->requireDependencyFilter && \count($matches) > 1) {
```
GromNaN, line 725: I removed this block in 6330be222cc3641e0f47cadd8af2609ec27500fb.

Really not understood what It was supposed to do. The packages does not have an `id` when they comes from the `RepositorySet` (line 710), not the `Pool` (line 713).

tests/PackageSelection/PackageSelectionTest.php
```diff
@@ -446,13 +471,13 @@ public function dataSelect(): array
                 $packages['alpha'],
                 $packages['epsilon'],
                 $packages['gamma1'],
+                $packages['gamma2'],
+                $packages['gamma3'],
```
GromNaN, line 475: I updated this test. `project-eta` requires `"vendor/project-gamma": ">=1"`. It seems obvious that all versions of `gamma` are eligible.

ihor-sviziev:
Hi @alcohol,
Could you review this PR? Would love to have composer 2 support in satis!

alcohol:
I can't really review it as I do not use Satis enough myself.

If community members can test this branch, then that would be greatly appreciated.

I have only one concern: will these changes impact Composer 1 users in a negative manner?

ralflang:
I have run this through some non-trivial real world use cases and it does work.

alcohol:
If @GromNaN can resolve the conflicts and considers it finished, I'll merge it.

GromNaN:
Branch has been rebased.
It works on my own satis repositories (they uses rather simple configurations).

alcohol:
Merged, though I also added PHPStan (at level 1) and it already found some issues that I think in part resulted from updating to Composer 2. Perhaps when you have time you can take a look and see if any of those are noteworthy and/or easily fixed?

GromNaN:
@jk if you are trying to use a pear repository. [This is no longer supported.](https://getcomposer.org/upgrade/UPGRADE-2.0.md)

Otherwise, could you open an issue with a reproductible case?

ihor-sviziev:
I think, in case if composer 1 is not supported - we might release these changes as the next major version. 
It will highlight that it's backward-incompatible, and in the readme/changelog file, we might highlight that Satis version 2.0 doesn't support composer 1.x anymore.

What do you think?

GromNaN:
A new major is on its way.
https://github.com/composer/satis/commit/263796ef40312f8ab9180f54c349073751c06d52#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34

alcohol:
`main` is always current. If you want to stick to the old release, you can switch to the new `2.x` branch.

PR #657, opened 2021-07-13
Add a minify option

Deamon:
To follow up #594 task list,
here is the PR to add the minify option.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -29,7 +29,7 @@ jobs:
         run: |
           composer install
           composer --working-dir=tools/php-cs-fixer install
-      - run: composer run php-cs-fixer
+      - run: composer run php-cs-fixer --diff
 
   phpstan:
     runs-on: ubuntu-latest
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -29,7 +29,7 @@ jobs:
         run: |
           composer install
           composer --working-dir=tools/php-cs-fixer install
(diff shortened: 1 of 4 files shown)


alcohol:
I see packagist only uses it in one place: https://github.com/composer/packagist/blob/ecd1d34940d6bb5ecc932abaae745d8d9e9b0c05/src/Package/V2Dumper.php#L252

Perhaps you could use this as a point of reference?

Deamon:
Thanks for the hint @alcohol! 
With it I was able to find the perfect place for the minify method.

alcohol:
> Thanks for the hint @alcohol!
> With it I was able to find the perfect place for the minify method.
> 
> Do you think I should add
> 
> ```
> "composer/metadata-minifier": "^1.0"
> ```
> 
> to composer.json now we're using it directly in the code ?

You're welcome! And thank you for taking the time to take a look at this issue  :-)

While the package is currently a transient requirement via `composer/composer` (which is one of our direct dependencies), it doesn't hurt to add it to our own list of direct requirements.

.github/workflows/lint.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: |
           composer install
           composer --working-dir=tools/php-cs-fixer install
-      - run: composer run php-cs-fixer
+      - run: composer run php-cs-fixer --diff
```
alcohol, line 32:
```suggestion
      - run: composer run php-cs-fixer
```

Deamon:
And thank you for your time!

PR #670, opened 2021-09-07
fixes the package filter

klj613:
similar to https://github.com/composer/satis/commit/7128ad33b2682a23e0ddc47f436e66cc74ecf7df
this seems to be a regression from when repositoryUrl was changed to `InputOption::VALUE_IS_ARRAY`
since when the repositoryUrl is not set its an empty array


alcohol:
Thanks!

PR #676, opened 2021-09-17
Add doc block to Application->getComposer()

pavog:
Copied from Composer\Factory::create()


SvenRtbg:
It's duplicating the existing docblock above.

pavog:
Whoops, seems like I mixed something up when copy pasting.

PR #683, opened 2021-10-13
Allow empty repo and other errors, when skipErrors is enabled

DeyV:


alcohol:
I'm not sure if these errors should be skipped even if skip errors is enabled, since they indicate that the user probably misconfigured something. Then again, since using the skip errors flag is a conscious decision, I guess it is at the user's discretion.

PR #684, opened 2021-10-20
Fix errors for phpstan level 6

### LINKED ISSUE #658: Increase PHPStan to level 6
Todo:

- [ ] Increase PHPStan configuration to level 6
- [ ] Resolve all issues

j3j5:
Hi, I tried to fix all errors from phpstan on level 6. Not sure if you have a specific wish for the format for iterable types (`array<int,string>` vs `string[]`) but for the rest it should be fine. 

Actually, I think I've found a bug on the implementation of strip-hosts, but please double check that part before merging just in case I misunderstood how it's supposed to work.

It should fix #658

### DIFF
--- a/src/Builder/ArchiveBuilderHelper.php
+++ b/src/Builder/ArchiveBuilderHelper.php
@@ -20,9 +20,15 @@ class ArchiveBuilderHelper
 {
     /** @var OutputInterface The output Interface. */
     private $output;
-    /** @var array The 'archive' part of a configuration file. */
+    /** @var array<string, mixed> The 'archive' part of a configuration file. */
     private $archiveConfig;
 
+    /**
+     *
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
+     * @param array<string, mixed> $archiveConfig
+     * @return void
+     */
     public function __construct(OutputInterface $output, array $archiveConfig)
     {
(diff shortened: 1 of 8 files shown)


src/Builder/ArchiveBuilderHelper.php
```diff
@@ -87,6 +99,12 @@ protected function isOneOfNamesInList(array $names, array $list): bool
         return false;
     }
 
+    /**
+     *
+     * @param string $name
```
GromNaN, line 104: Some `@param` and `@return` annotations are redundant with method signature; is it really required by phpstan?

alcohol: I don't believe so.

These was a [discussion about the superfluous phpdoc rule](https://github.com/composer/satis/discussions/679); but so far consensus was to keep the rule in place.

dereuromark: It is also more consistent for the time being.
And to be honest:
Removing some and adding others is usually counterproductive in readability, it really confuses people as to the positioning of those lines then if they ever need to be added.

alcohol:
That merge commit was not what I intended to do. Let me see if I can resolve this from CLI.

alcohol:
See #731

PR #688, opened 2022-03-12
Update to Bootstrap 5, fix prettier ci check

theimbender:
- Updates all npm dependencies
- Makes changes to the scss files to work with Bootstrap 5
- Ran prettier on files so that calls to the `prettier:ci` script should pass


alcohol:
Cheers

Issue #690, opened 2022-03-18
Latest update to composer.lock breaks composer install on PHP 7.4


smottt:
**Describe the bug**

The latest update to `composer.lock` locks `psr/log` to `2.0.0` which bumps the php requirement to 8.0+. But according to [composer.json](https://github.com/composer/satis/blob/main/composer.json#L33), PHP 7.3+ is required. If I remove `composer.lock`, everything installs perfectly fine.

Commit: https://github.com/composer/satis/commit/75b29ca57d65f95e94b0f4eb704ad9e461eb4729#diff-f37acfaa6b11f575a9a6f41a75fa73a61d0f8ebc2c9b8cddc215d8aca10e44f5R676

**To Reproduce**
Run the following with PHP 7.4:

```
composer create-project composer/satis:dev-main
```

**Outcome**

```
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - psr/log is locked to version 2.0.0 and an update of this package was not requested.
    - psr/log 2.0.0 requires php >=8.0.0 -> your php version (7.4.28) does not satisfy that requirement.
  Problem 2
    - symfony/deprecation-contracts is locked to version v3.0.0 and an update of this package was not requested.
    - symfony/deprecation-contracts v3.0.0 requires php >=8.0.2 -> your php version (7.4.28) does not satisfy that requirement.
  Problem 3
    - symfony/filesystem is locked to version v6.0.6 and an update of this package was not requested.
    - symfony/filesystem v6.0.6 requires php >=8.0.2 -> your php version (7.4.28) does not satisfy that requirement.
  Problem 4
    - symfony/finder is locked to version v6.0.3 and an update of this package was not requested.
    - symfony/finder v6.0.3 requires php >=8.0.2 -> your php version (7.4.28) does not satisfy that requirement.
  Problem 5
    - symfony/mime is locked to version v6.0.3 and an update of this package was not requested.
    - symfony/mime v6.0.3 requires php >=8.0.2 -> your php version (7.4.28) does not satisfy that requirement.
  Problem 6
    - symfony/process is locked to version v6.0.5 and an update of this package was not requested.
    - symfony/process v6.0.5 requires php >=8.0.2 -> your php version (7.4.28) does not satisfy that requirement.
  Problem 7
    - symfony/service-contracts is locked to version v3.0.0 and an update of this package was not requested.
[...]

PR #691, opened 2022-03-20
Clean up .prettierignore, update dependencies

theimbender:
- Clean up the entries in `.prettierignore` so that only generated files should be ignored
- Update npm/composer dependencies to latest
- Remove `.gitignore` from `.husky` directory, that's not needed in latest version of husky
- Updated readme link that still pointed to license file on master branch


theimbender:
Should we be running `composer update` on php 8.1 or 7.3? I was running it on 7.3 and got some downgraded dependencies, isn't that desirable since it's the minimum supported version?

theimbender:
Seems to be the same issues as #690, I think we'd want to only commit lock files generated with the minimum supported version of php

alcohol:
Yeah I am not sure. Personally I would rather only support the latest versions of PHP. But Satis is a community product. If the community feels a need to support older version(s), then we should pin the PHP configuration to the lowest supported version in the `composer.json` file. That way, no mater what environment you run `composer update` on, the configured PHP version will be leading.

theimbender:
This should be good to merge now, I think, now that the composer lock file has been rebased

PR #692, opened 2022-03-23
Composer's Package class functions now require hash maps of Links rather than just arrays

theimbender:
This should hopefully fix the CI test failures, fixes phpstan level 5 errors


alcohol:
Cheers

PR #696, opened 2022-04-15
Updated the Symfony/Console version

hughcube:


SvenRtbg:
You cannot simply change the `composer.json` without changing the `composer.lock` file. If you do, it would invalidate the lock file checksum at least.

Issue #700, opened 2022-08-17
RFC: Satis as a composer 2.3 plugin (POC implementation available)


ralflang:
# Running Satis as a Composer Plugin

tldr - I have mostly implemented a composer plugin that makes a slightly patched satis work as a composer 2.3.x plugin. I want to know if anybody is interested in this before I spend more time on it.

## Motivation

I use satis in the build pipeline for a 100+ packages framework. I have recently begun to expand the framework's composer installer plugin to replace older standalone tools that are currently used in CI jobs and in the development process. Discovering the similarities between satis and composer I decided to spend two evening hours on making satis run inside composer

## State of the PoC

All four commands are wrapped into shim classes which changes the name to satis:add, satis:build etc and nicely groups them below the native composer commands in a special satis section.

The add and purge commands work as normal. 
All commandline options are passed work as desired. I did only limited testing.
The build command sort of works:
- packagist is not used unless explicitly configured in satis.json
- in require-all mode, all packages of the configured repos are downloaded. If you enable the packagist repo, you will essentially get a local. static mirror
- commandline repo filters do work
- Explicit require lists in the satis.json file currently don't work as expected. Instead, as soon as the satis file contains a require section, the require section of the root package is used. In standalone mode, satis abuses a composer installation and pretends its config file is a root package. I will need to trick the already running composer into accepting satis' config file as the root package.
- commandline package filters were not tested but likely are affected by the root package issue.

- In plugin mode, satis does not install its own composer/composer version (unless it is the root package) but runs against the composer phar file which calls into it.

## Potential uses/benefits

Composer has a more mature but similar base application, event system, and plugin ecosystem. Hooking satis into composer allows us to shrink satis' codebase. Developers / volunteers can focus on actual functionality and leave infrastucture work to composer/composer which is already maintained much better. Third party plugin developers can leverage satis for their automation or integration plugins and call into it by emitting events (provided satis moves into that direction).

One could easily integrate the little boilerplate plugin into satis itself, in other words make satis itself a composer plugin. I decided to leave that unexplored for the moment and gather feedback first. Maybe there is no interest in this at all.

As a side effect I fixed some incompatibilities between satis (dev-main) and composer 2.3.x 

## Questions

- Is there any interest at all in this? If not, at least it was a fun experiment.
- Which line of development would you prefer, making satis itself the plugin or making satis only capable of working with the plugin library and otherwise leaving it as it is?

ralflang:
[...]

Issue #694, opened 2022-03-30
Satis no longer working with Composer 2.3?


BSchwetzel:
**Describe the bug**
I believe satis is no longer working with Composer 2.3:

**To Reproduce**
Try a satis build with Composer 2.3

**Outcome**
```
$ php satis/bin/satis build -vvv satis.json /var/jenkins_home/composer-repo

In BaseCommand.php line 59:
[RuntimeException]                                                           
Composer commands can only work with an Composer\Console\Application instance set

Exception trace:
  at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/vendor/composer/composer/src/Composer/Command/BaseCommand.php:59
 Composer\Command\BaseCommand->getApplication() at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/src/Console/Command/BuildCommand.php:183
 Composer\Satis\Console\Command\BuildCommand->execute() at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/vendor/symfony/console/Command/Command.php:298
 Symfony\Component\Console\Command\Command->run() at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/vendor/symfony/console/Application.php:1015
 Symfony\Component\Console\Application->doRunCommand() at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/vendor/symfony/console/Application.php:299
 Symfony\Component\Console\Application->doRun() at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/src/Console/Application.php:49
 Composer\Satis\Console\Application->doRun() at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/vendor/symfony/console/Application.php:171
 Symfony\Component\Console\Application->run() at /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/satis/bin/satis:60

build [--repository-url REPOSITORY-URL] [--repository-strict] [--no-html-output] [--skip-errors] [--stats] [--minify] [--] [<file> [<output-dir> [<packages>...]]]
```

**Expected behavior**
```
$ php satis/bin/satis build -vvv satis.json /var/jenkins_home/composer-repo

Loading config file /var/jenkins_home/jobs/XXX/jobs/satis-update/workspace/auth.json
Checked CA file /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a file.
[...]

PR #701, opened 2022-08-18
Make satis compatible with composer 2.3, 2.4 and add plugin mode

ralflang:
- Fix compatibility issues with composer 2.3/2.4 in CLI mode
- Integrate ralflang/satis-plugin natively
- Change app kernel into a child of composer's app kernel and remove parts better inherited
- Satis as a root project will work as before, auto-installs a version of composer/composer for autoloading reasons
- Satis as a non-root project will give guidance that composer/composer is no longer automatically installed and what to do
- updated phpstan tool to not spam the CI log with php notices
- Updated unit tests to be phpstan l5 compliant after the composer/composer upgrade created breakage due to nested array type annotations

adresses #700, #694

### DIFF
--- a/bin/satis
+++ b/bin/satis
@@ -22,6 +22,17 @@ if ((!$loader = includeIfExists(__DIR__.'/../vendor/autoload.php')) && (!$loader
     exit(1);
 }
 
+if (!class_exists('Composer\Console\Application')) {
+    print("The Satis standalone CLI requires composer. If satis is not your root project, run\n
+    composer require --dev composer/composer\n\nin your root project or use the equivalent plugin commands:\n
+    composer satis:init
+    composer satis:build
+    composer satis:add
+    composer satis:purge
+    \n");
+    exit(2);
+}
+
 if (function_exists('ini_set')) {
     // Set user defined memory limit or use Composer's user defined memory limit
     if ($memoryLimit = getenv('SATIS_MEMORY_LIMIT')) {
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
     },
     "require": {
         "php": "^7.4||^8.0",
-        "composer": "^2.3.10",
+        "composer/composer": "^2.3.10",
         "composer-plugin-api": "^2.3.0",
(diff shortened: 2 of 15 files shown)


alcohol:
Ah, sorry, can you fix the conflicts first?

ralflang:
Yes, I have merged the conflicts. Meanwhile PHPStan has some complaints and I think a re-test with different composer versions is due after work today. I will highlight you once I am done with that.

alcohol:
PHPStan passing isn't really a requirement, more of a guide / nice-to-have. But I'm fine with waiting till after your test session(s) before merging this PR.

ralflang:
@alcohol Test went through

In Standalone mode with "lowest"  and "latest" dependencies allowed by constraints:
```
./bin/satis 
./bin/satis init
./bin/satis add  --type vcs --name "horde/text_diff" https://github.com/maintaina-com/text_diff
./bin/satis build satis.json /tmp/composer/
```
Lowest gives one deprecation warning in a dependency.

In plugin mode (only tested with latest dependencies)
```
composer satis:init
composer satis:add --type vcs --name "horde/text_diff" https://github.com/maintaina-com/text_diff
composer satis:build ./satis.json /tmp/composer/
```

I think we are good to go.

There are two places which smell like they might sense to be changed elsewhere.
For the sake of getting done I worked around them for now.

Composer/Application constructor does not accept optional name and version parameters.
This makes sense from Composer's POV as it thinks it is an app while satis CLI treats it as a library.
https://github.com/composer/satis/pull/701/files#diff-c3161949b1e6a9be81ac8e805ddd3d9f9a8da07e10e3932c297dfadf1ea47da7R37

If leaving out optional arguments to RootPackageLoader, it constructs its own version of VersionGuesser and in turn ProcessExecutor which it cannot really use.
https://github.com/composer/satis/pull/701/files#diff-1c8b1c04ba573f2bb91f1cfa0ce3bb2eca35a8004c5caa12abbf51d86496589dR203

Do you think it makes sense to open issues in the appropriate repos for further exploring these?

Issue #708, opened 2022-10-12
Change in PackageSelection repository filters


MaximilianKresse:
The last commit broke our satis build process. This change here (from @ralflang) - to be specific: https://github.com/composer/satis/commit/bf82c45f818b5061e17be0eeaf5664c02d370274#diff-d16a5020dc605f5cb12e0cf898bb6a7d2b2a7eacd2dc2bbc77b56b1d61d005a3L169

This version now also includes all packages from packagist in the RepositoryManager (if packagist isn't disabled in the [new required composer.json](https://github.com/composer/satis/issues/707)). 
This is probably the reason why this default repo filter was added in the first place.

The problem now is we have currently multiple external package lists like this:
```
{
    "repositories": [
        {
            "type": "composer",
            "url": "https://www.setasign.com/downloads/"
        }
    ]
}
```

Previously all packages from this repository were included in the package list. But the newly added default repo filter currently removes these.

Everything works like before if I disable packagist with a composer.json like this:
```
{
    "name": "satis/test",
    "type": "library",
    "require": {},
    
    "repositories": [
        {
            "packagist.org": false
        }
    ]
}
```
And remove the newly added default repo filter (the added else branch from L176 - L194).

This feature is an absolute MUST-HAVE for us (and probably many others too). Was this change indended and do you have any ideas to solve this? A setting to disable the default repo filter would be sufficient IMO.

ralflang:
Sure, I will look into why the filter does not apply anymore. This is not intended.

MaximilianKresse:
It does apply. But the problem is there wasn't a filter before and the newly added default filter now only include repositories which are directly added to the list of repositories. But before I used "type": "composer" to load another list without having to care which packages are inside this repository.

With this change I would have to build my satis.json configuration dynamically.

ralflang:
I have a hard time understanding some details. This is probably on my side.

The intended behaviour is as before:
- no composer.json file required
- packagist is off unless it's explicitly on

Can you please provide me the satis.json that worked until yesterday (or an anonymized version thereof)

My best guess is it's something like this:
```
{
    "name": "foo.com/foo-repo",
    "homepage": "https://foo.com",

    "require-all": true,
    "repositories": [
        {"name": "something/orother", "url": "https://github.com/maintaina-com/components", "type": "composer"},
        ...
    ]
}
```

MaximilianKresse:
> The intended behaviour is as before:
> 
> * no composer.json file required
[...]

Issue #707, opened 2022-10-12
"Composer could not find a composer.json file"


MaximilianKresse:
**Describe the bug**
When using satis in cli mode it currently throws an error:

```
Composer could not find a composer.json file in /var/www/webs/test
To initialize a project, please create a composer.json file. See https://getcomposer.org/basic-usage
```

The given directory is the current working directory.

**To Reproduce**
```
mkdir test && cd test
php8.0 /path/to/satis/bin/satis init
php8.0 /path/to/satis/bin/satis build satis.json ./build
```

The problem is solved if I just create an empty composer.json like this:
```
{
    "name": "satis/test",
    "type": "library",
    "require": {}
}

```

**Expected behavior**
Previously no composer.json was required. Is this an intended change?

jeanfrauc:
@MaximilianKresse Check the message "_The Satis standalone CLI requires composer. If satis is not your root project, run composer require --dev composer/composer in your root project_" they added in the `bin/satis` file when they changed the project satis to `composer-plugin` in their composer.json.
https://github.com/composer/satis/pull/701/files
Adding the command `composer require --dev composer/composer` to my CICD project worked for me without having to add a composer.json file myself.

MaximilianKresse:
@jeanfrauc I'm a little bit confused about your answer - where do you get this error? This message doesn't pop up for me. I have simply cloned this project and used "composer update". In the default "require" is already ["composer/composer": "^2.3.10"](https://github.com/composer/satis/blob/main/composer.json#L46) so I don't see any sense to adding composer as an additional dev requirement.

I don't have a "CICD project". I had simply called "php8.0 /path/to/satis/bin/satis build satis.json /var/www/web/satis" to build my satis packages. Now I had to create a composer.json next to the satis.json.

ralflang:
Yes, I have an idea where this comes from. A patch will be provided. As a workaround until tomorrow you may want to roll back your satis version to before that commit.

jeanfrauc:
OK, sorry @MaximilianKresse if it wasn't clear enough.

In fact, I had exactly the same error message as you ("Composer could not find a composer.json file...") when I was building my packages.
But I am not cloning the satis project tu build my packages as you do. I am using their Docker image (which is called "composer/satis").
My CICD project is simply a Jenkinsfile in which I pull the Docker image and then I do a "docker run" with several parameters to call the satis build command.

[...]

PR #709, opened 2022-10-13
Make standalone mode run without a composer.json file

ralflang:
Address composer/satis#707 composer/satis#708 
- Fix composer in standalone mode trying to load a composer.json & failing
- Fix including packagist by default
- Fix reading repos of type composer


chrfritsch:
I can confirm that the issue I have with latest:

```
Scanning packages
In PackageSelection.php line 185:
                             
  Undefined array key "url"  
```

Is fixed with this PR. 👍🏼

PR #716, opened 2023-03-13
fix: change abandonned box compactors

JamesRezo:
using Box version 4.3.7, the compactor classes must be changed


alcohol:
For reference: https://github.com/box-project/box/commit/d58544b9500c85e9b64a7a88ea1ca564379ada5a

PR #730, opened 2023-09-09
Update front end dependencies, purge unused CSS from bundle

theimbender:
Makes the following changes to the web view:
- Update all dependencies, adding some new ones as required in the migration guides of the existing dependencies (`@symfony/webpack-encore` [now requires](https://github.com/symfony/webpack-encore/releases/tag/v4.0.0) several other webpack packages, for example)
- Update the version of node in the CI config from 15 to to 20, and add an `.nvmrc` file to encourage development on node 20 as well. Node 20 is about to become the [only supported version of node](https://nodejs.dev/en/about/releases/).
- Add [PurgeCSS](https://purgecss.com/) webpack plugin to encore to remove unused CSS, reducing size of uncompressed CSS bundle nearly 80% from 108kb to 22kb
- Added ECMAScript 2017 to parser option for eslint, since prettier now likes to add trailing commas


alcohol:
Thanks!

PR #733, opened 2023-09-16
Increase PHPStan rule level to 7

### LINKED ISSUE #659: Increase PHPStan to level 7
Todo:

- [ ] Increase PHPStan configuration to level 7
- [ ] Resolve all issues

theimbender:
Resolves #659 by increasing PHPStan rule level to 7 to account for possible issues with union types. I think I've followed the logic correctly, but could definitely use feedback on whether there should be additional exceptions or if there are better ways of handling these.


src/Builder/ArchiveBuilder.php
```diff
@@ -119,7 +119,8 @@ public function dump(array $packages): void
                 $distUrl = sprintf('%s/%s/%s/%s', $endpoint, $this->config['archive']['directory'], $intermediatePath, $archive);
                 $package->setDistType($archiveFormat);
                 $package->setDistUrl($distUrl);
-                $package->setDistSha1Checksum($includeArchiveChecksum ? hash_file('sha1', $path) : null);
+                $hasedPath = $includeArchiveChecksum ? hash_file('sha1', $path) : false;
+                $package->setDistSha1Checksum(false !== $hasedPath ? $hasedPath : null);
```
GromNaN, line 123: `hash_file` will never return a falsy string (`''` or `'0'`)
```suggestion
                $package->setDistSha1Checksum($includeArchiveChecksum ? (hash_file('sha1', $path) ?: null) : null);
```

src/PackageSelection/PackageSelection.php
```diff
@@ -289,13 +296,13 @@ public function load(): array
         }
 
         if (isset($rootConfig['providers']) && is_array($rootConfig['providers']) && isset($rootConfig['providers-url'])) {
-            $baseUrl = $this->homepage ? parse_url(rtrim($this->homepage, '/'), PHP_URL_PATH) . '/' : '';
+            $baseUrl = is_string($this->homepage) ? parse_url(rtrim($this->homepage, '/'), PHP_URL_PATH) . '/' : '';
             $baseUrlLength = strlen($baseUrl);
 
             foreach ($rootConfig['providers'] as $package => $provider) {
-                $file = str_replace(['%package%', '%hash%'], [$package, $provider['sha256']], $rootConfig['providers-url']);
+                $file = (string) str_replace(['%package%', '%hash%'], [$package, $provider['sha256']], $rootConfig['providers-url']);
```
GromNaN, line 303: If you enforce the type of the input, the type of `str_replace` output should be correct.
```suggestion
                $file = str_replace(['%package%', '%hash%'], [$package, $provider['sha256']], (string) $rootConfig['providers-url']);
```

src/PackageSelection/PackageSelection.php
```diff
@@ -577,15 +584,17 @@ private function matchAddr($addr1, $addr2, $len = 0, $chunklen = 32): bool
...
      * @throws \Exception
      */
     private function addRepositories(RepositorySet $repositorySet, array $repositories): void
     {
         foreach ($repositories as $repository) {
             try {
-                $repositorySet->addRepository($repository);
+                if ($repository instanceof RepositoryInterface) {
+                    $repositorySet->addRepository($repository);
+                }
```
GromNaN, line 597: This looks like an actuel bugfix. But I don't think this is possible.

src/PackageSelection/PackageSelection.php
```diff
@@ -577,15 +584,17 @@ private function matchAddr($addr1, $addr2, $len = 0, $chunklen = 32): bool
     }
 
     /**
-     * @param RepositoryInterface[] $repositories
+     * @param array<RepositoryInterface|ConfigurableRepositoryInterface> $repositories
```
GromNaN, line 587: It should be a union type. I don't see any class that implement `ConfigurableRepositoryInterface` without `RepositoryInterface`.

```suggestion
     * @param array<RepositoryInterface&ConfigurableRepositoryInterface> $repositories
```

src/PackageSelection/PackageSelection.php
```diff
@@ -689,15 +698,15 @@ private function getFilteredLinks(PartialComposer $composer): array
...
+                return in_array($link->getTarget(), $packagesFilter, true);
             }
         );
 
         return array_values($links);
     }
 
     /**
-     * @param RepositoryInterface[] $repositories
+     * @param array<RepositoryInterface|ConfigurableRepositoryInterface> $repositories
```
GromNaN, line 709:
```suggestion
     * @param array<RepositoryInterface&ConfigurableRepositoryInterface> $repositories
```

src/Builder/ArchiveBuilder.php
theimbender: The return type for `hash_file` is `string|false` and this rule level disallows short ternary operators, I'll tweak this a little though.

src/PackageSelection/PackageSelection.php
theimbender: I don't know if we want to change this from union to intersection types, when I do that there are several branches that would always evaluate to true or become unreachable, and some methods that use one of those for a parameter type start throwing errors as well.

GromNaN: Ok then, feel free to mark as resolved.

src/Builder/ArchiveBuilder.php
```diff
@@ -119,7 +119,8 @@ public function dump(array $packages): void
                 $distUrl = sprintf('%s/%s/%s/%s', $endpoint, $this->config['archive']['directory'], $intermediatePath, $archive);
                 $package->setDistType($archiveFormat);
                 $package->setDistUrl($distUrl);
-                $package->setDistSha1Checksum($includeArchiveChecksum ? hash_file('sha1', $path) : null);
+                $hasedPath = hash_file('sha1', $path);
+                $package->setDistSha1Checksum($includeArchiveChecksum ? (is_string($hasedPath) ? $hasedPath : null) : null);
```
alcohol, line 123: Nitpicking, but I think "hased" is a typo?

PR #735, opened 2023-09-24
Update front end accessibility

theimbender:
This update provides a few tweaks to the front end view that should get the Lighthouse accessibility score to 100:
- Update NPM dependencies
- Include all of Bootstrap instead of just trying to comment out what's not used. Since PurgeCSS was added to the project to remove unused styles in #730, we don't have to do that manually and it just makes it harder if you're trying to use parts of Bootstrap that would need to be re-added
- Change how the collapsible sections work, since I had it backwards when I first implemented ARIA. This change puts `aria-expanded` on the control rather than container element, and adds `aria-controls` to support the relationship between control and container.
- Use the [`hidden`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden) property to prevent collapsed sections from being tabbed to
- Add a debounced listener to window `resize` event to update the heights of collapsed sections on window resizes
- Add `aria-label` attribute to `select` elements with no associated `label`
- Update some CSS class names, Bootstrap has classes that specifically add good text color contrast that we should be taking advantage of
- Update the link icon next to each package's title to be the same color as the link text so that it's visible


alcohol:
Thanks!

PR #736, opened 2023-09-26
Allow generated html output to be indexed by search engines

theimbender:
This change allows satis to achieve a 100% Lighthouse score for SEO by adding a new config option to not block search engine indexing, `allow-seo-indexing`. This option is `false` by default, to preserve existing behavior.

We are also adding a meta description to the generated html output. It will try to use the config file's description if set, or a fallback description if not.


alcohol:
Thanks!

PR #909, opened 2024-09-01
Fix php 8.1

Giacomo92:
I have this problem with satisfy
Screenshot 2024-09-01 alle 22 55 50

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "ext-zip": "*",
         "composer/composer": "^2.7.7",
         "composer-plugin-api": "^2.3.0",
-        "symfony/console": "^7.1.3",
+        "symfony/console": "^6.0",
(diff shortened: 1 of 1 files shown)


SvenRtbg:
In fact if the project requires "php >= 8.1", it should allow dependencies compatible with 8.1. Still I would think that a higher minor version of 6 should be used (6.0 requires PHP >= 8.0.2), and in addition I would keep the 7.x version for those that can make use of it.

alcohol:
Resolved with [2df2549](https://github.com/composer/satis/commit/2df25498b62080d0ab088347e2b63f671056eae8), though I imagine the resolution will not please you. Alas, since this is a (barely maintained) community project, we want to keep the BC promises to an absolute minimum. We only strive to support the latest of everything.

PR #920, opened 2024-09-26
Add 'includes' option to satis schema

### LINKED ISSUE #751: `includes` configuration is documented but not valid
**Describe the bug**
https://github.com/composer/satis/blob/main/docs/config.md#includes says `includes` can be configured.

https://github.com/composer/satis/blob/main/res/satis-schema.json does not have it.

**To Reproduce**
```
{
    "name": "drupal/test",
    "homepage": "http://test.example.org",
    "repositories": [
        { "type": "vcs", "url": "https://github.com/drupal/core-composer-scaffold.git" },
        { "type": "vcs", "url": "https://github.com/drupal/core-project-message.git" },
        { "type": "vcs", "url": "https://github.com/drupal/core.git" },
        { "type": "vcs", "url": "https://github.com/drupal/core-recommended.git" }
    ],
    "require-all": true,
    "includes": true,
    "archive": {
        "directory": "dist",
        "prefix-url": "http://test.example.org"
    }
}
```

`satis build --minify satis.json output`

**Outcome**

Attempting to use it errors with

```
The property includes is not defined and the definition does not allow additional properties

In BuildCommand.php line 348:

  The json config file does not match the expected JSON schema
```

**Expected behavior**

I do not want to support Composer 1.

onspli:
PR #726 added an option to disable output of all Composer v1 metadata with `includes` option. The option has never been added to `satis-schema.json` so an attempt to actually use the option cases error 
```
The property includes is not defined and the definition does not allow additional properties
```
Fixes #751

### DIFF
--- a/res/satis-schema.json
+++ b/res/satis-schema.json
@@ -195,6 +195,11 @@
             "description": "Specify filename instead of default include/all$%hash%.json",
             "default": "include/all$%hash%.json"
         },
+        "includes": {
+            "type": "boolean",
+            "description": "If true, output package includes. This is `true` by default - setting it to `false` allows you to work with Composer v2 metadata URLs only.",
+            "default": true
+        },
         "available-package-patterns": {
             "type": "array",
             "description": "Composer v2 feature. List of patterns like 'vendor/*' for packages available, recommended with many packages. If not set, 'available-packages' will be set with ALL package names.",


onspli:
Fixes #751

alcohol:
Thanks

PR #1104, opened 2026-03-03
Various PHPStan fixes

theimbender:
This pull request makes the following small fixes:

1. Fixes the `ArchiveBuilderTest.php` test because the anonymous class extending `ArchiveManager` was not calling the parent constructor.
2. Removes a baseline from `phpstan-baseline.neon` that hasn't been needed in a while and was breaking expectations. Not sure if the baseline was ever actually being applied because of how anonymous class names are generated.
3. Updates the package name in `package-lock.json`, it probably got clobbered during a dependency upgrade.

These changes should allow the `phpstan` action to succeed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-    "name": "tmp.Xjy0jDX5Qw",
+    "name": "satis",
     "lockfileVersion": 3,
     "requires": true,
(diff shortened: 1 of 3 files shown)


alcohol:
Thanks 🥳

PR #1153, opened 2026-06-15
bump minimum PHP version to 8.4

loru88:
the composer.json claimed PHP 8.3 is supported but when I used that in our private corporate registry, it failed because the recently updated dependencies require PHP 8.4, hence I thought to bump the minimum version to PHP 8.4 and update the composer packages

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [8.3, 8.4, 8.5]
+        php: [8.4, 8.5]
         stability: [prefer-lowest, prefer-stable]
     name: PHP ${{ matrix.php }} - ${{ matrix.stability }}
     steps:
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN set -eux ; \
     --no-suggest \
     --classmap-authoritative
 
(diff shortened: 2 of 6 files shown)


alcohol:
Could probably also bump all the other constraints in the composer.json while you're at it. And the workflows for 8.3 and lower (if any) can be dropped.

loru88:
> Could probably also bump all the other constraints in the composer.json while you're at it. And the workflows for 8.3 and lower (if any) can be dropped.

I already did `composer update`.
I  pushed a new commit to also update the tools composer packages, did you mean that?

alcohol:
@loru88 Yeah but I meant `composer bump`; which updates all the constraints in `composer.json` to match the current installed version or higher.

Also; I meant removing 8.3 from the list of versions in https://github.com/composer/satis/blob/main/.github/workflows/ci.yaml#L19

alcohol:
Not sure about the Dockerfile change; that actually downgrades us from 8.5 to 8.4.x - is that intentional?

loru88:
yes I did it on purpose to avoid the issue from happening again.
Fixing the PHP version to the same version expected by composer it should ensure the compatible packages and dependencies, otherwise it might install packages compatible with PHP 8.5 but not with PHP 8.4 again.

in a next iteration we could bump to PHP 8.5 maybe

PR #1197, opened 2026-08-27
Add a remove command to drop a repository from the Satis config

### LINKED ISSUE #1064: composer satis:remove
Hello,

Please add the command `satis:remove`

Examples: 
```bash
composer satis:remove git@github.com:user/project.git
php bin/satis remove git@github.com:user/project.git
```

ousamabenyounes:
Fixes #1064

## Summary

`add` puts a repository URL into `satis.json`, but nothing takes one back out —
the only way to drop a repository today is to hand-edit the JSON. This adds the
mirror image of `AddCommand`:

    php bin/satis remove git@example.com:user/project.git
    composer satis:remove git@example.com:user/project.git

Both invocation forms from the issue work: the command is registered in
`Console\Application::getDefaultCommands()` for the standalone binary and in
`Plugin\CommandProvider` as `satis:remove` for the Composer CLI.

Scope follows the reading in the issue thread — the opposite of `AddCommand`,
i.e. it removes a repository entry from the Satis config. It does not touch
`require`, the built repository, or the archives; `build` and `purge` already
own those, and `add` does not touch them either.

## Behaviour

It reuses `AddCommand`'s argument shape and exit codes, so the two commands stay
symmetric:

| Situation | `add` | `remove` |
|---|---|---|
| config path is an http(s) URL | `2` | `2` |
| config file does not exist | `1` | `1` |
| URL already in / not in the config | `4` (already there) | `4` (not there) |
| success | `0` | `0` |

Details worth calling out:

- Every entry sharing the URL is removed, and the list is reindexed with
  `array_values()` so `repositories` stays a JSON array rather than turning into
  an object with numeric string keys.
- Entries without a `url` key (a `package` repository, for instance) are never
  removed.
- Removing needs no network access — unlike adding, there is no repository to
  validate — so the command is fully offline and unit-testable.

## Test verification (RED → GREEN)

RED — the new `RemoveCommandTest` plus the updated `CommandProviderTest`
expectation, applied to unmodified `main` with no production change
(`src/Console/Command/` contains only Add/Build/Init/Purge):

```
PHPUnit 11.5.55 by Sebastian Bergmann and contributors.

CommandProvider
 ✔ Implements the CommandProviderCapability interface
 ✘ Returns the expected commands in order
   │ Failed asserting that actual size 4 matches expected size 5.
   │ tests/Plugin/CommandProviderTest.php:45

RemoveCommand
 ✘ Rejects an HTTP URL as the config file path
   │ Error: Class "Composer\Satis\Console\Command\RemoveCommand" not found
 ✘ Rejects a repository URL that is not in the config
   │ Error: Class "Composer\Satis\Console\Command\RemoveCommand" not found
 ✘ Removes the only repository from the config
   │ Error: Class "Composer\Satis\Console\Command\RemoveCommand" not found
 ... (11 errors in total)

ERRORS!
Tests: 13, Assertions: 2, Errors: 11, Failures: 1.
```

GREEN — same tests, with the command:

```
PHPUnit 11.5.55 by Sebastian Bergmann and contributors.

CommandProvider
 ✔ Implements the CommandProviderCapability interface
 ✔ Returns the expected commands in order

RemoveCommand
 ✔ Rejects an HTTP URL as the config file path
 ✔ Rejects an HTTPS URL as the config file path
 ✔ Rejects a config file that does not exist
 ✔ Rejects a repository URL that is not in the config
 ✔ Rejects a repository URL when the config has no repositories at all
 ✔ Rejects a repository URL when the repositories value is not an array
 ✔ Removes the only repository from the config
 ✔ Preserves the other repositories and reindexes the list
 ✔ Removes every entry sharing the same URL
 ✔ Leaves the rest of the configuration untouched
 ✔ Keeps repository entries that have no url

OK (13 tests, 47 assertions)
```

Also exercised through the real binary, on the SCP-style VCS URL form the
issue uses:

```
--- satis.json before ---
{
  "name": "demo/repo",
  "repositories": [
    {"type": "vcs", "url": "git@example.com:user/project.git"},
    {"type": "vcs", "url": "git@example.com:user/keep.git"}
  ]
}

$ php bin/satis remove git@example.com:user/project.git /tmp/satis.json

                                                                                      
  Your configuration file successfully updated! It's time to rebuild your repository  
                                                                                      

exit=0
--- satis.json after ---
{
  "name": "demo/repo",
  "repositories": [
    {
      "type": "vcs",
      "url": "git@example.com:user/keep.git"
    }
  ]
}

$ php bin/satis remove git@example.com:user/project.git /tmp/satis.json   # again
Repository url not found in the file
exit=4
```

## Full suite

`composer run test`, before and after, on every matrix cell of `ci.yaml`
(PHP 8.4 / 8.5 × `prefer-lowest` / `prefer-stable`):

| | Tests | Assertions | Incomplete | Failures |
|---|---|---|---|---|
| `main` | 172 | 317 | 5 | 0 |
| this branch | 183 | 354 | 5 | 0 |

`php-cs-fixer`, `phpstan` (level 8 + strict + deprecation rules), `eslint`,
`prettier` and `yamllint` were also run before and after: green in both cases,
with no new entry in `phpstan-baseline.neon`.

No documentation file changes, because the repo does not enumerate its CLI
commands anywhere — `add` and `init` are not mentioned in the README or under
`docs/` either. `remove` documents itself through `setHelp()`, exactly like
`add`. Happy to add a README section covering `add` and `remove` together if
you would like one.

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -69,6 +69,7 @@ class Application extends ComposerApplication
             new Command\AddCommand(),
             new Command\BuildCommand(),
             new Command\PurgeCommand(),
+            new Command\RemoveCommand(),
         ]);
 
         return $commands;
--- a/src/Console/Command/RemoveCommand.php
+++ b/src/Console/Command/RemoveCommand.php
@@ -1,0 +1,99 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of composer/satis.
+ *
+ * (c) Composer <https://github.com/composer>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace Composer\Satis\Console\Command;
+
+use Composer\Command\BaseCommand;
+use Composer\Json\JsonFile;
+use Symfony\Component\Console\Helper\FormatterHelper;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class RemoveCommand extends BaseCommand
+{
+    protected function configure(): void
+    {
+        $this->getName() ?? $this->setName('remove');
+        $this
+            ->setDescription('Remove repository URL from satis JSON file')
+            ->setDefinition([
+                new InputArgument('url', InputArgument::REQUIRED, 'VCS repository URL'),
+                new InputArgument('file', InputArgument::OPTIONAL, 'JSON file to use', './satis.json'),
+            ])
+            ->setHelp(
+                <<<'EOT'
+                The <info>remove</info> command removes the given repository URL from the json
+                file (satis.json is used by default). You will need to run <comment>build</comment> command
+                to regenerate your repository without that repository, and <comment>purge</comment> to
+                drop the archives it left behind.
+                EOT
+            );
+    }
+
+    /**
+     * Exit codes mirror AddCommand: 2 for a remote config file, 1 for a config
+     * file that does not exist, 0 on success. Where AddCommand returns 4 for a
+     * URL that is already in the config, this returns 4 for a URL that is not.
+     */
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        /** @var FormatterHelper $formatter */
+        $formatter = $this->getHelper('formatter');
+
+        $configFile = $input->getArgument('file');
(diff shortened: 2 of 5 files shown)


alcohol:
The documentation in `docs/` is a long outstanding issue (see #403), but not critical. Most users of Satis know how it works or seem able to figure it out regardless. If you have time, you are more than welcome to update them.

PR #1202, opened 2026-09-09
Register the repositories of the config only once per build

philstutz:
The repositories of a Satis config reach the `RepositoryManager` three times: through `Factory::create($io, $config)` in standalone mode (`src/Console/Application.php:82`), through the loop in `BuildCommand::execute()` (`BuildCommand.php:227`), and through `RootPackageLoader::load($config)` (`BuildCommand.php:241`). `PackageSelection` scans every duplicate on its own, so each repository is fetched and read three times per build — three `git remote update --prune origin`, three `git remote show origin` and three composer.json reads per tag and branch. On our instance (436 VCS repositories) a full build takes ~4 h instead of ~1 h.

Reproducer, empty `COMPOSER_HOME`, `satis --no-ansi -v build satis.json out`:

```json
{
    "name": "example/duplicate-scan-repro",
    "homepage": "https://example.org",
    "repositories": [
        { "type": "git", "url": "https://github.com/Seldaek/signal-handler.git" },
        { "type": "git", "url": "https://github.com/composer/spdx-licenses.git" }
    ],
    "require-all": true,
    "output-html": false
}
```

| `Reading composer.json of …`     | before | after |
| -------------------------------- | ------ | ----- |
| composer/spdx-licenses (24 refs) | 72     | 24    |
| seld/signal-handler (8 refs)     | 24     | 8     |

Resulting package set unchanged.

This PR drops the duplicates from the manager once all three registrations happened, keeping the first occurrence and comparing `ConfigurableRepositoryInterface` repositories by type and URL — same `ReflectionProperty` approach `removeDisabledRepositories()` already uses. The loop stays, plugin mode relies on it; suppressing the re-add inside `RootPackageLoader::load()` is not possible from Satis.

Tests: `BuildCommandTest` gains three cases, one of them replaying the real registration sequence (3 before, 1 after). 191 tests green, PHPStan and php-cs-fixer clean.

Present since #701 and #709 (2022-10).

### DIFF
--- a/src/Console/Command/BuildCommand.php
+++ b/src/Console/Command/BuildCommand.php
@@ -241,6 +241,7 @@ class BuildCommand extends BaseCommand
         $loader = new RootPackageLoader($manager, $composerConfig, $parser, $guesser);
         $satisConfigAsRootPackage = $loader->load($config);
         $composer->setPackage($satisConfigAsRootPackage);
+        $this->deduplicateRepositories($manager, $output, (bool) $verbose);
 
         $packageSelection = new PackageSelection($output, $outputDir, $config, $skipErrors);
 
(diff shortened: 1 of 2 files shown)


philstutz:
@alcohol Thanks!

alcohol:
No, thank you 👍

